PR #378, opened 2020-02-19
fix: reconfigure graphql server for monorepo

vuhuucuong:
Reconfigure `graphql-sever` for` monorepo` compatible reason & tweak some other configs

To run the server:
- `lerna boostrap`
- `lerna run build:prod --scope @reapit/elements --scope @reapit/cognito-auth --stream --parallel` first since `graphql-server` depends on those two packages.
By default `lerna (yarn)` will use local version if `package.json -> version` of local packages are satisfied, otherwise it will try to install from `npm registry`).
- Then just `yarn workspace graphql-server start:dev`


packages/graphql-server/tsconfig.json
```diff
@@ -1,19 +1,25 @@
...
-    "paths": {
-      "@/*": ["src/*"],
-      "@reapit/cognito-auth": ["../cognito-auth/src"],
-      "@reapit/elements": ["../elements/src"]
-    }
+    "module": "commonjs",
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "target": "es6",
+    "jsx": "react",
```
phmngocnghia: This package is now pointing to the production version of @reapit/elements which doesn't contain JSX code. I think this option seems to be redundant

Issue #373, opened 2020-02-19
Fix resolvers for contact identity after break changes


tanphamhaiduong:
![image](https://user-images.githubusercontent.com/22660833/74815748-623c3200-532c-11ea-8b3c-4815d7958a4c.png)

Rename folder contact-identity to identityChecks
Doc here: https://dev.marketplace.reapit.cloud/developer/swagger

PR #391, opened 2020-02-21
feat: #373 add more resolver for identity services

tanphamhaiduong:
fixes: #373


phmngocnghia:
api.ts should be tested

packages/graphql-server/src/resolvers/identity-check/identity-check.d.ts
```diff
@@ -0,0 +1,61 @@
+export type Document = {
```
willmcvay: These definitions are in TS Definitions right? Bit confused why we are copy-pasting them

packages/graphql-server/src/resolvers/identity-check/resolvers.ts
```diff
@@ -0,0 +1,51 @@
...
+import errors from '../../errors'
+import identityCheckServices from '../identity-check/services'
+import {
+  CreateIdentityCheckArgs,
+  UpdateIdentityCheckArgs,
+  GetIdentityCheckByIdArgs,
+  GetIdentityChecksArgs,
+} from './identity-check'
+
+export const queryIdentityChecks = (_: any, args: GetIdentityChecksArgs, context: ServerContext) => {
```
willmcvay: Again, can you add return types

packages/graphql-server/src/resolvers/identity-check/services.ts
```diff
@@ -0,0 +1,51 @@
...
+import logger from '../../logger'
+import { ServerContext } from '../../app'
+import {
+  callCreateIdentityCheckAPI,
+  callUpdateIdentityCheckAPI,
+  callGetIdentityCheckByIdAPI,
+  callGetIdentityChecksAPI,
+} from './api'
+
+export const getIdentityCheckById = (args: GetIdentityCheckByIdArgs, context: ServerContext) => {
```
willmcvay: And again, return types

packages/graphql-server/src/resolvers/identity-check/identity-check.d.ts
tanphamhaiduong: So we decide to maintain GraphQL server definition separately. In the future and we can map params on GraphQL server if services change any params name, we do not change to avoid breaking FE and it'll take more effort if we change schema and need to fix FE

Issue #369, opened 2020-02-19
Resolver for Graphql-server to call Area Services


tanphamhaiduong:
![image](https://user-images.githubusercontent.com/22660833/74815441-d0342980-532b-11ea-881a-e3c3d0a4fcd6.png)

We need to write resolvers for area services in `graphql-server`

Doc here: https://dev.marketplace.reapit.cloud/developer/swagger

PR #419, opened 2020-02-24
feat: #369 GraphQL Server Area resolver

vuhuucuong:
fixes: #369

### DIFF
--- a/packages/graphql-server/src/resolvers/area/__tests__/resolvers.test.ts
+++ b/packages/graphql-server/src/resolvers/area/__tests__/resolvers.test.ts
@@ -1,18 +1,81 @@
 import areaServices from '../services'
 import { checkPermission } from '../../../utils/check-permission'
-import logger from '../../logger'
-import errors from '../../errors'
-import { ServerContext } from '../../app'
-
+import errors from '../../../errors'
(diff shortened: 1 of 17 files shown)


packages/graphql-server/src/resolvers/area/area.d.ts
```diff
@@ -0,0 +1,85 @@
+import { AuthenticationError, UserInputError } from 'apollo-server'
+
+export interface AreaModel {
```
willmcvay: These types need to come from Definitions

packages/graphql-server/src/resolvers/area/__tests__/resolvers.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  generateAuthenticationError: jest.fn(() => 'authentication error'),
+}))
+jest.mock('../../../logger')
+jest.mock('../../../utils/check-permission', () => ({
+  checkPermission: jest.fn(() => true),
+}))
+
+describe('queryArea', () => {
+  it('should return correctly', () => {
+    ;(checkPermission as any).mockReturnValue(true)
```
willmcvay: Just a small point buy the type here is jest.Mock rather than any, I don't mind too much as it's a test

PR #428, opened 2020-02-24
chore: #377 show validation errors on Submit/Update page

### LINKED ISSUE #377: Show validation errors on the bottom of the Submit/Edit app pages
**Summary:** A developer will be notified a validation has occurred but will not know which field it refers to unless they scroll up and view that particular section. This doesn’t provide a very good user experience. We should show all validation errors in the white space currently occupied by the ‘Submit App’ button. 

**Task:** To display the validation errors to the developer at the bottom of the ‘Submit’ and ‘Edit’ pages in the developer portal. 

**For example:** 
This is the following error that occurs when the characters in the summary box exceeds the limit 

image

We should display this (in addition the message that appears around the field) to the developer at the bottom of the page.
- **Heading:** The following validation errors have occurred:
- Should include the field label and the description of the error

See example: 
The following validation errors have occurred:
Summary: The value provided exceeds the maximum length of 150 characters

See mockup: 
image

See mockup of multiple validation errors: 
image

phmngocnghia:
fixes #377


packages/marketplace/src/components/pages/developer-submit-app.tsx
```diff
@@ -75,6 +76,46 @@ export interface SubmitAppMappedProps {
...
+  if (isErrorsEmpty) {
+    return null
+  }
+
+  const description = errors[FIELD_ERROR_DESCRIPTION]
+
+  return (
+    <div className="has-text-danger">
+      <H6 className="has-text-danger mb-1">
+        {description ? `${description}:` : 'The following validation errors have occurred:'}
```
vuhuucuong: I think we can use 
```
        {description  ||  'The following validation errors have occurred:'}
```
instead

Issue #374, opened 2020-02-19
Deploy Graphql server to lambda


tanphamhaiduong:
Deploy GraphQL server to lambda

tanphamhaiduong:
can access by graphiQL app by click on https://8g8ibl8iei.execute-api.eu-west-2.amazonaws.com/dev/graphql

PR #435, opened 2020-02-25
chore: #374 config to deploy graphql-server to lambda

tanphamhaiduong:
fixes: #374


packages/graphql-server/package.json
```diff
@@ -9,35 +9,57 @@
...
+    "color-name": "^1.1.4",
+    "colors": "^1.4.0",
     "dataloader": "^2.0.0",
     "express": "^4.17.1",
     "graphql": "^14.5.8",
     "graphql-depth-limit": "^1.1.0",
     "graphql-import": "^0.7.1",
     "graphql-type-json": "^0.3.1",
+    "is-stream": "^2.0.0",
+    "isomorphic-fetch": "^2.2.1",
```
willmcvay: Are all of these dependencies or do some really live in dev?

Also, we already have iso-fetch right?

willmcvay: Am fine with us hoisting the dependencies because it's quite a special service however, I think I copied a lot of them to the root so there is probably some duplication where we can now remove stuff from the root package json where it is only used in GQL server

willmcvay: Not a big deal, just some housekeeping!

tanphamhaiduong: @willmcvay let me try to remove it and use from the root packages

tanphamhaiduong: let me try to remote isomophic fetch and use from root package.json

packages/graphql-server/webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+    ],
+  },
+  plugins: [
+    new CopyPlugin([
+      {
+        test: /\.(graphql|gql)$/,
+        ignore: ['*.ts', '*.test.ts', 'tests/**/*'],
+        from: 'src',
+        to: 'src',
+        force: true,
```
vuhuucuong: Not sure what this means 🤔 
```
        from: 'src',
        to: 'src',
```
Everything else LGTM 👍

Issue #424, opened 2020-02-24
Marketplace not loading in Internet Explorer 


HollyJoyPhillips:
**Summary:** Unable to load the Marketplace on IE 11. We need to test the functionality of the Marketplace in IE but cannot get the login page to load. 

**Browser Version:** V 11.418.18362.0 

**Location:** https://dev.marketplace.reapit.cloud/ 

**See error in console:** 
![image](https://user-images.githubusercontent.com/52408607/75147608-9ed5a800-56f5-11ea-8475-53578a06b149.png)

PR #446, opened 2020-02-26
Fix: #424 - Remove query-string lib to fix IE11 compatibility issues

nphivu414:


package.json
```diff
@@ -195,7 +195,7 @@
     "purgecss": "^1.3.0",
     "purgecss-whitelister": "^2.4.0",
     "purify-css": "^1.2.5",
-    "query-string": "^6.9.0",
+    "query-string": "5.0.0",
```
nphivu414: Need to downgrade `query-string` to v5 because v6 only targets the latest version of Chrome, Firefox, and Safari.

query-string installation guide:
"This module targets Node.js 6 or later and the latest version of Chrome, Firefox, and Safari. If you want support for older browsers..."
https://github.com/sindresorhus/query-string#install
https://github.com/sindresorhus/query-string/issues/189

packages/marketplace/package.json
```diff
@@ -8,7 +8,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "build:prod": "node '../../scripts/webpack/webpack-sass-prod/index.js'",
+    "build:prod": "node \"../../scripts/webpack/webpack-sass-prod/index.js\"",
```
nphivu414: Have to use `" "` here to run this script on Windows machine

vuhuucuong: you can remove the quote here since it's not necessary
```
"build:prod": "node ../../scripts/webpack/webpack-sass-prod/index.js",
```

package.json
willmcvay: Can we not just get rid of the lib and use URLSearchParams - not supported in IE but babel will just polyfill it

PR #455, opened 2020-02-27
fix: swagger curl issue

### LINKED ISSUE #438: Interactive API requests don't work
**Summary:**
The API console doesn't display request details properly

**Steps to reproduce:**
Submit any API request through the interactive API console. Let's take, for example, the applicant view.

**Expected Results:**
A curl command for the API request and applicant details.

**Actual Results:**
`curl -X "undefined"` and `TypeError: Failed to fetch`
Screenshot from 2020-02-25 17-58-13

willmcvay:
Fixes #438


packages/marketplace/src/components/pages/swagger.tsx
```diff
@@ -10,35 +10,48 @@ import { StringMap } from '../../types/core'
...
 }
 
 export const SwaggerPage: React.SFC = () => {
   const [loading, setLoading] = React.useState(true)
+  const [accessToken, setAccessToken] = React.useState()
+  const requestInterceptor = (params: StringMap) => fetchInterceptor(params, accessToken)
+
+  React.useEffect(() => {
+    fetchAccessToken(setAccessToken)
+  })
```
vuhuucuong: a small thing, do we need `[fetchAccessToken]` as a dependency here to avoid `fetchAccessToken` from being called again and again?

willmcvay: So fetchAccessToken calls getAccessToken which contains all the logic that checks if the token is valid / has expired etc.
It is not sufficient to check if the token exists (because it could exist but be invalid), so I call out to the method which performs the check on validity of the token and just returns it if it's valid so it's not expensive.
If it has expired however, it will refresh my session etc

Issue #450, opened 2020-02-26
Implement platform version to GraphQL service


willmcvay:
Outcome of #420 - we have gone with option 2 ie endpoint versioning v1, v2 etc. Option 1 had merits but would be very difficult to mix and match versions, also, would be confusing for clients when using TS Definitions, even potentially impossible with mis-matched resources.

- Add `yarn upgrade-platform-version` (or similar), command to package.json
- Task will do 2 things - set the API to a defined date (for platform versions), and to a specific version of TS-Defintions.
- First task to set the API-Version constant that is used in all headers that call to the platform so all API calls are all set to a given platform version. The current version is  2020-01-31 and we should use this as the initial base.
- Second task - Checked with Craig and the Swagger doc is not versioned so we will need to fetch the correct types for this date by checking out the correct tag for ts-defintions release on foundations eg https://github.com/reapit/foundations/tree/foundations-ts-definitions_v0.0.52/packages/foundations-ts-definitions/types
- Will save these to local directory, local tsconfig to point to local definitions when referencing `@reapit/foundations-ts-definitions` in the code.
- Each of the PRs currently, should then delete any local type definitions and refer to `@reapit/foundations-ts-definitions`
- We will manually upgrade to each platform version when we are ready as we move towards v1 public Alpha version.
- When we have a stable beta we will freeze breaking changes and make LTS v1 of the service, patches and non-breaking changes only.
- v2 will then be what we develop against for new features, v1 will slowly be deprecated with client notices to upgrade to the new LTS.

PR #470, opened 2020-03-02
feat: #450 create fetch-ts-definition command for graphql server

trankhacvy:


packages/graphql-server/scripts/fetch-ts-definitions.js
```diff
@@ -0,0 +1,71 @@
...
+const fs = require('fs')
+
+const ITEM_PER_PAGE = 100
+
+const files = [
+  { src: 'packages/foundations-ts-definitions/types/index.ts', dest: 'src/types/index.ts' },
+  { src: 'packages/foundations-ts-definitions/types/platform-schema.ts', dest: 'src/types/platform-schema.ts' },
+]
+
+const fetchTSDefinitions = async () => {
```
tanphamhaiduong: pls add try catch

Issue #66, opened 2020-01-22
As a 'Developer', I should be able to Query and Mutate Offices entity using GraphQL


HollyJoyPhillips:
**Summary:** See above, the offices entity should be made available in the Apollo GQL Schema. 

**Task:** To write in Apollo Server package

- Query for all Offices
- Query for office by id
- Mutation for new Office 
- Mutation for office by id

PR #511, opened 2020-03-03
feat: #66 Offices page on SMB

trankhacvy:


packages/smb/src/components/pages/offices/index.js
```diff
@@ -0,0 +1,3 @@
+import Offices from './offices'
```
willmcvay: Should be index.ts not index.js right?

PR #521, opened 2020-03-04
fix: fix failing pipeline

vuhuucuong:
- Fix webpack not throwing errors when compile failed in some cases
- `yarn build` now run in order instead of parallel
- Temporary disable sentry source map since it doesn't work and causing compilation error
- Add explicit dependencies to each package, so `lerna/yarn workspace` knows how to build package in the correct order

Note:
I think it's still not ideal since we have to manually set alias to `@reapit/elements` & `@reapit/cognito-auth` in every `tsconfig.json`. It might have some other ways to automatically resolve those packages to their correct location in monorepo by config our `webpack-config/tsconfig` .
But for now to unblock everyone, just temporary use that. We can find a better solution later 🙂


packages/aml-checklist/package.json
```diff
@@ -22,6 +22,8 @@
     "lint": "concurrently \"tsc --noEmit\" \"eslint --cache --ext=ts,tsx,js src\""
   },
   "dependencies": {
+    "@reapit/elements": "^0.5.39",
```
tanphamhaiduong: why do we need to add this ?

packages/cognito-auth/package.json
```diff
@@ -30,6 +30,7 @@
     "lint": "concurrently \"tsc --noEmit\" \"eslint --cache --ext=ts,tsx,js src\""
   },
   "dependencies": {
+    "@reapit/elements": "^0.5.39",
```
tanphamhaiduong: same as above

scripts/webpack/webpack.base.prod.js
```diff
@@ -74,12 +74,12 @@ module.exports = {
       cacheId: process.cwd(),
       cleanupOutdatedCaches: true,
     }),
-    new SentryWebpackPlugin({
-      include: '.',
-      ignoreFile: '.sentrycliignore',
-      ignore: ['node_modules', 'webpack.config.js'],
-      configFile: 'sentry.properties',
-    }),
+    // new SentryWebpackPlugin({
```
tanphamhaiduong: pls add TODO to this comment

packages/aml-checklist/package.json
vuhuucuong: We need this so `yarn build` know which package to build first.
For example:
`aml-checklist` depends on `@reapit/elements` & `@reapit/cognito-auth`. and `@reapit/cognito-auth` depends on `@@reapit/elements`. Then `lerna/yarn workspace` will build `@reapit/elements` first -> then `@reapit/cognito-auth` -> `aml-checklist`

vuhuucuong: Example from material-ui: https://github.com/mui-org/material-ui/blob/master/packages/material-ui/package.json#L49

tanphamhaiduong: @vuhuucuong could we move this to root elements instead of let it on each packages ?

vuhuucuong: I think no, because if we move it to root `package.json`, then `lerna/yarn workspace` cannot know which package depends on which package -> then `yarn build` will build in the wrong order -> and cause our pipeline to fail again
For now, I'm finding a way to config our `webpack/tsconfig` so  that we don't need to set paths alias in tsconfig.json in each package anymore or use one `tsconfig.json` for alias all packages like this
https://github.com/mui-org/material-ui/blob/master/tsconfig.json

Issue #199, opened 2020-02-07
Setup authentication for the Search Widget 


HollyJoyPhillips:
- Machine to machine authentication and API handling

Issue #127, opened 2020-01-29
Web Components need to be lightweight as possible. Investigate and prototype solutions.


willmcvay:
Our first web component version, the Search Widget is a React App. This made sense when we were doing rapid prototyping however, for first public Beta they should be framework agnostic and super lightweight.

With this in mind we should investigate both Polymer and Svelte as possible solutions as lightweight standalone components.

Outcome of this ticket is to demonstrate a working MVP of the search widget

PR #559, opened 2020-03-09
feat: web components svelte refactor

willmcvay:
Fix: #127 
Fix: #199


packages/web-components/src/search-widget/client/api/constants.ts
```diff
@@ -0,0 +1,15 @@
+const CONSTANTS: { [key: string]: { [key: string]: string } } = {
+  LOCAL: {
+    WEB_COMPONENT_API_BASE_URL_SEARCH_WIDGET: 'http://localhost:3000',
+  },
+  DEV: {
+    WEB_COMPONENT_API_BASE_URL_SEARCH_WIDGET: 'http://localhost:3000',
+  },
+  PROD: {
+    WEB_COMPONENT_API_BASE_URL_SEARCH_WIDGET: 'http://localhost:3000',
```
tanphamhaiduong: So will we change it to ENV later right ?

packages/web-components/src/search-widget/client/api/constants.ts
```diff
@@ -0,0 +1,15 @@
+const CONSTANTS: { [key: string]: { [key: string]: string } } = {
+  LOCAL: {
+    WEB_COMPONENT_API_BASE_URL_SEARCH_WIDGET: 'http://localhost:3000',
+  },
+  DEV: {
+    WEB_COMPONENT_API_BASE_URL_SEARCH_WIDGET: 'http://localhost:3000',
```
tanphamhaiduong: So will we change it to ENV later right ?

packages/web-components/package.json
```diff
@@ -9,53 +9,42 @@
...
+    "serverless-offline": "^5.12.1",
+    "serverless-webpack": "^5.3.1",
+    "sirv-cli": "^0.4.5",
+    "svelte-jester": "^1.0.4",
+    "svelte-preprocess": "^3.4.0"
   },
-  "peerDependencies": {
-    "react": "^16.10.2",
-    "react-dom": "^16.10.2"
+  "dependencies": {
```
vuhuucuong: you should include `@reapit/cognito-auth` as a dependency here, so yarn build can build in correct order, and also if we want to use a fixed version of it, just provide the version number in here.

vuhuucuong: another cool thing about this is : hen you make some changes in `@reapit/cognito-auth`, all the tests in all packages which have `@reapit/cognito-auth` as a dependency will also be run by `lerna run test:update-badges`. This make sure our changes in `@reapit/cognito-auth` doesn't break another apps.
The same thing's applied for `@reapit/elements`, which forces you to update snapshot of all apps after making changes in `@reapit/elements`. That's very useful :)

willmcvay: Cognito Auth is not a dependency of this package and won't be - this is client credentials auth flow so the client side app does not need the user to login

packages/web-components/src/search-widget/client/api/constants.ts
willmcvay: Updated

packages/web-components/src/search-widget/client/api/constants.ts
willmcvay: Done!

PR #565, opened 2020-03-10
fix: Update the return data of update mutations

trankhacvy:


vuhuucuong:
Can you please run `yarn workspace build:schema` to generate `schema.graphql` file and include that file in commit

packages/graphql-server/src/resolvers/appointments/api.ts
```diff
@@ -109,7 +109,20 @@ export const callUpdateAppointmentAPI = async (
       },
       body: payload,
     })
-    return response
+    if (updateResponse) {
+      const getResponse = await fetcher({
```
tanphamhaiduong: pls call getById instead of create new fetcher in this

packages/graphql-server/src/resolvers/areas/api.ts
```diff
@@ -97,7 +97,20 @@ export const callUpdateAreaAPI = async (args: UpdateAreaArgs, context: ServerCon
       },
       body: payload,
     })
-    return response
+    if (updateResponse) {
+      const getResponse = await fetcher({
```
tanphamhaiduong: pls call getById instead of create new fetcher in this

vuhuucuong:
LGTM!

Issue #73, opened 2020-01-22
Create Query to populate 'Users' table in SMB App


HollyJoyPhillips:
**Relates to ticket:** https://github.com/reapit/foundations/issues/71

**Summary:** Create Apollo server side query to populate the 'Users' table on the SMB App

- Query Apollo service to get Negotiators by company id 
- Should also support upload of offices by CSV 
- Table should populate with correct data

PR #570, opened 2020-03-10
#73: Create query to populate users table smb

nphivu414:


tanphamhaiduong:
I think should use negotiators term instead of the user, it'll cause confusion when we read the code

PR #571, opened 2020-03-10
fix: spreadsheet component fix afterDataChanged callback

vuhuucuong:
- Added diffing algorithm to calculate `changedCells` when data is changed
- `afterDataChanged` callback now receives `(data: Cell[][], changedCells)` as arguments
- `validate` is now a prop of `Spreadsheet`, receive a function (like Formik)


packages/elements/src/components/Spreadsheet/utils.tsx
```diff
@@ -51,6 +64,40 @@ export const convertToCompatibleData = (parsedResult): Cell[][] =>
 /**
  * Convert back from Cell[][] to string[][]
  */
-export const convertDataToCsv = (data: Cell[][]): string[][] => data.map(rowArray => rowArray.map(({ value }) => value))
+export const convertDataToCsv = (data: Cell[][]): string[][] =>
+  data.map(rowArray => rowArray.map(({ value }) => value)) as string[][]
 
-export const hideContextMenu = (prev: ContextMenuProp): ContextMenuProp => ({ ...prev, visible: false })
+// Diffing algorithm to find differences between data array
+export const changedCellsGenerate = (newData?: Cell[][], oldData?: Cell[][]): ChangedCells => {
```
willmcvay: Worth keeping an eye on how this performs - loops within loops can be really heavy. Might need to user a worker if it gets slow...

vuhuucuong: I will keep an eye on it, maybe we can implement a better algorithm or using Web Worker as you said sounds cool.

Issue #567, opened 2020-03-10
Create mutations should return new object instead of true or false


tanphamhaiduong:
Context:
Create mutations for entities now just return true or false after create
For the new id, we can get from the headers after create

Output:
Create mutations should return new object instead of true or false

tanphamhaiduong:
![image](https://user-images.githubusercontent.com/22660833/76301678-3a1f6d80-62f1-11ea-80cd-6263233a4d6f.png)
@willmcvay there is no create id for negotiators from headers after created, could you help me to check with backend guys. I'll move this ticket back to backlog

Issue #549, opened 2020-03-06
GraphQL Server should throw correct error messages


vuhuucuong:
### Summary:
Currently, our `GraphQL Server` only has two error types `AuthenticationError` & `UserInputError`.
![Screen Shot 2020-03-06 at 21 13 33](https://user-images.githubusercontent.com/26746668/76091217-ba8f5700-5fef-11ea-8b29-c184e9c2c0bb.png)
We should add more error types, based on the API responses of *Platform API Server*

### Task:
- Return the correct error message:
  - If it's `Platform API` error, pass the error message to `Client`
  - If it's an error specific to GraphQL server, define the corresponding error message and send it to `Client`
- Also fix the `AuthenticationError` since it seems to not work as expected.

PR #604, opened 2020-03-12
feat: #567 create mutation should return object instead of true or false and add handleError to return correctly

tanphamhaiduong:
Relate #567 #549 
Changes:
Migrate fetcher to axios to request to other services
Add handleError function for reusable

### DIFF
--- a/packages/graphql-server/nodemon.json
+++ b/packages/graphql-server/nodemon.json
@@ -2,5 +2,5 @@
   "watch": ["src"],
   "ext": "ts,ejs,graphql",
   "ignore": ["src/**/*.test.ts", "src/schema.graphql"],
-  "exec": "yarn build:schema && ts-node -r tsconfig-paths/register -P tsconfig.json ./src/app.ts"
+  "exec": "ts-node -r tsconfig-paths/register -P tsconfig.json ./src/app.ts"
 }
--- a/packages/graphql-server/src/errors/errors.ts
+++ b/packages/graphql-server/src/errors/errors.ts
@@ -5,7 +5,7 @@ export const errorMessages = {
   notAuthorized: '[E4010] Not Authorized',
   badRequestError: '[E4000] Bad Request',
   notFound: '[E4040] Not Found',
-  forBidden: '[E4030] Forbidden',
+  forbidden: '[E4030] Forbidden',
   internalServerError: '[E5000] Internal Server Error',
 }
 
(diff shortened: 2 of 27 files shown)


packages/graphql-server/nodemon.json
```diff
@@ -2,5 +2,5 @@
   "watch": ["src"],
   "ext": "ts,ejs,graphql",
   "ignore": ["src/**/*.test.ts", "src/schema.graphql"],
-  "exec": "yarn build:schema && ts-node -r tsconfig-paths/register -P tsconfig.json ./src/app.ts"
+  "exec": "ts-node -r tsconfig-paths/register -P tsconfig.json ./src/app.ts"
```
vuhuucuong: If you remove `yarn build:schema` here, can you change the `start:dev` to 
`cross-env REAPIT_ENV=LOCAL NODE_ENV=development yarn build:schema && nodemon`
in `package.json`, so it only builds one time, before start the dev server

vuhuucuong:
Another thing, Could we remove `@reapit/elements` from `package.json` & `tsconfig.json` now?

tanphamhaiduong:
> Another thing, Could we remove `@reapit/elements` from `package.json` & `tsconfig.json` of `graphql-server` for now?

Fixed as comment

packages/graphql-server/src/errors/errors.ts
```diff
@@ -19,14 +16,42 @@ export const generateAuthenticationError = (traceId?: string) => {
...
+export const generateInternalServerError = (traceId?: string) => {
+  const error = new ApolloError(`${traceId || ''} - ${errorMessages.internalServer}`, 'INTERNAL_SERVER_ERROR')
+  logger.info('generateInternalServerError', { traceId, error: JSON.stringify(error) })
+  return error
+}
+
+export const generateNotFoundError = (traceId?: string) => {
+  const error = new ApolloError(`${traceId || ''} - ${errorMessages.notFound}`, 'NOT_FOUND')
+  logger.info('generateNotFoundError', { traceId, error: JSON.stringify(error) })
+  return error
```
vuhuucuong: I think we should `throw error` here instead of return, like this: 
Screen Shot 2020-03-12 at 15 11 32

So that the `try/catch` will work correctly, and the `thrown errors` will be auto handled correctly by `apollo`

vuhuucuong: Same for above errors

tanphamhaiduong: @vuhuucuong actually I re-map errors to avoid show stack trace to user

tanphamhaiduong: image
this one is return error on the stack trace
image
because I use this function to remap and avoid show stack trace to the user

PR #624, opened 2020-03-16
fix: #601 invalidate chunk when deploying to dev/prod

### LINKED ISSUE #601: Deploying to dev / prod invalidates chunks
See errors like this in Sentry: 
https://sentry.io/organizations/reapit-ltd/issues/1525063499/?referrer=slack
https://sentry.io/organizations/reapit-ltd/issues/1557757085/?referrer=slack

To replicate: 
- Create a session on any page of marketplace and leave in your browser.
- Merge a PR and wait for deployment to dev
- Navigate to another page without refreshing

Screen will go blank as router cannot find next chunk - have to do a full refresh to fix.

vuhuucuong:
fixes #601 
We're having exactly the same issue as this: [Webpack deployments with code-splitting](https://spectrum.chat/react/general/how-to-handle-deployments-with-webpack-code-splitting~16dc5f85-0b4b-4eee-8693-066d82dcdc7f). Basically, this is what happened:
- Users are using the old version of `marketplace`.
- We merge code into master -> trigger DEV deployment.
- Old AWS S3 bucket is deleted -> all `.js, .css` files are deleted.
- Users request for the old `.js` chunks -> couldn't find because they were deleted.

As suggested in the above link, we can use webpack [offline-plugin](https://github.com/NekR/offline-plugin) to solve this. The plugin also provides a better UX since all the chunks will get pre-downloaded in background to browser cache, so that users can navigate between routes much faster


tanphamhaiduong:
@vuhuucuong actually we can make service worker run under the hood to announce user there is new version and advise them to refresh page

Issue #629, opened 2020-03-16
Config manager should not reference local config


willmcvay:
The current config manager update function makes it too easy to override the remote config. 

Currently;
- Fetches remote config, copies local config, then merges and pushes. This means a stale local config can override a newer remote.

Should;
- Fetch remote config, user should make their changes manually to this file and push. 

Or;
- Better still, diff changes and give a command prompt for each change asking the user to confirm each one is correct. This should prevent accidental overrides.

tanphamhaiduong:
I think we should separate local, production and, dev environment to 3 secrets management and we should update on AWS instead of run CLI

PR #644, opened 2020-03-18
chore: #629 migrate config of apps to config.json

tanphamhaiduong:
Changes:
- Add config.example.json
- Add method to fetch config.json from public folder for webpack
- Change `process.env. ...` to `window.reapit.config. ...`


packages/marketplace/src/types/global.d.ts
```diff
@@ -0,0 +1,20 @@
+export type Config = {
+  appEnv: string
```
willmcvay: Should be a type like 'local' | 'dev' | 'prod' etc right?

packages/marketplace/src/components/hocs/error-boundary.tsx
```diff
@@ -40,7 +40,8 @@ export class ErrorBoundary extends React.Component<ErrorProps, ErrorState> {
       type: 'COMPONENT',
       message: errorMessages.DEFAULT_COMPONENT_ERROR,
     })
-    if (process.env.NODE_ENV === 'production') {
+    const isDevelopment = window.reapit.config.appEnv === 'development'
```
willmcvay: So what is the app env for local development? I mean, sometimes we need to differentiate between local and development?

packages/marketplace/src/components/pages/login.tsx
```diff
@@ -48,8 +48,7 @@ export const tabConfigs = ({ loginType, history }: LoginProps): TabConfig[] => [
 ]
 
 export const Login: React.FunctionComponent<LoginProps> = (props: LoginProps) => {
-  const reapitEnv = process.env.REAPIT_ENV || 'LOCAL'
-  const isReapitEnvProd = reapitEnv === 'PROD'
+  const isProduction = window.reapit.config.appEnv === 'production'
```
willmcvay: In this case we need to differentiate between dev and prod - only prod hides the login options tabs

packages/marketplace/src/core/router.tsx
```diff
@@ -32,9 +32,9 @@ const RegisterConfirm = React.lazy(() => import('../components/pages/register-co
 const AdminStats = React.lazy(() => import('../components/pages/admin-stats'))
 
 const Router = () => {
-  const isReapitEnvProd = process.env.REAPIT_ENV === 'PROD'
+  const isDevelopment = window.reapit.config.appEnv === 'development'
```
willmcvay: What about local dev?

tanphamhaiduong: I think local dev env is the same with development environment so I let 2 things same

packages/marketplace/src/components/pages/login.tsx
tanphamhaiduong: yes only production hide login option tabs

PR #646, opened 2020-03-18
fix: #597 updates to desktop interaction for cloud apps

### LINKED ISSUE #597: Updates to Desktop Interaction for Cloud Apps
Background: 

For our November launch we demoed some interactions between our core CRM (Agency Cloud), and web apps in the Marketplace. Because of the changes required to authentication with Reapit Connect, this behaviour is now broken when receiving data from the CRM (sending data to the CRM works fine).

Previously:

The apps received a refresh token in the URL as a query `desktopToken="some-token"` and an optional list of params passed to the app from desktop. The desktopToken did two things -firstly, it allowed the app to refresh it's session and secondly, it set a value in the Redux `RefreshSession` called mode - by default this is "WEB" but was set to "DESKTOP" if the token was present.

If the "mode" was set to "DESKTOP" it allowed us to change the urls of our DynamicLinks (see Elements https://reapit.github.io/elements/?path=/story/dynamiclinks--dynamicbuttonsandlinks), and hide the sidebar of the app. 

Finally in some cases (the live example is in AML App), I want to ask Agency Cloud to return me a value. I do this by adding an `appParam` value to my DynamicLink so that agency cloud will return this value to me as a query string in "DESKTOP" mode. See:

```
{
    dynamicLinkParams: {
      appMode: 'DESKTOP',
      entityType: EntityType.CONTACT,
      entityCode: '',
      queryParams: {
        name: 'Bob',
        appParam: AppParams.CONTACT_CODE,
        appId: '516060af-ea0a-4384-8465-28befc6f34b3',
      },
    }
```
In this example, Agency cloud will do a search on a CONTACT name with string "Bob" then return me an id in the with the query `cntCode` (CONTACT_CODE), to the app by id (AML).

Task:

The behaviour should remain the same however, rather than the `desktopToken` the apps should rely on a global variable called `window.__REAPIT_MARKETPLACE_GLOBALS__` - this is an object, generated by AgencyCloud with the following function:

```
const setMarketplaceGlobals = (globals = {}) => {
  if (window.__REAPIT_MARKETPLACE_GLOBALS__) {
    window.__REAPIT_MARKETPLACE_GLOBALS__ = {
      ...window.__REAPIT_MARKETPLACE_GLOBALS__,
      ...globals
    }
  } else {
    Object.defineProperty(window, '__REAPIT_MARKETPLACE_GLOBALS__', {
      value: globals,
    })    
  }
}

```
* If the object exists, the `refreshSession.mode` should be set to "DESKTOP", else "WEB".
* If the `refreshSession.mode` is "DESKTOP", hide the sidebar menu. This needs to be applied to SMB, Marketplace, GeoDiary and AML
* If the object has values, I should be able to get them by key. Create a helper in Elements > DynamicLinks to do this.
* In AML Search, I will need to get the `cntCode` I currently reference in `client-search` in desktop mode from this object, not from the query string, using the helper a the previous step,

vuhuucuong:
fixes #597


packages/geo-diary/src/core/private-route-wrapper.tsx
```diff
@@ -36,9 +37,11 @@ export const PrivateRouteWrapper: React.FunctionComponent<PrivateRouteWrapperPro
   hasSession,
 }) => {
   const cognitoClientId = process.env.COGNITO_CLIENT_ID_GEO_DIARY as string
-  const refreshParams = getTokenFromQueryString(location.search, cognitoClientId)
+  const refreshParamsRaw = getTokenFromQueryString(location.search, cognitoClientId)
+  const marketplaceGlobalObject = getMarketplaceGlobalsByKey()
 
-  if (refreshParams && !hasSession) {
+  if (refreshParamsRaw && !hasSession) {
+    const refreshParams: RefreshParams = { ...refreshParamsRaw, mode: marketplaceGlobalObject ? 'DESKTOP' : 'WEB' }
```
willmcvay: Just a suggestion but would it not make more sense to add this to getTokenFromQueryString? Then you don't have to repeat yourself. Just a thought...

vuhuucuong: @willmcvay ah great, will check and change if it doesn't affect other things

willmcvay: It’s fine as long as you don’t change the returned interface. In this case mode is already returned so will be fine

vuhuucuong: Cool, thanks 💯

Issue #596, opened 2020-03-11
‘Completed’ requirements need amending in the AML app 


HollyJoyPhillips:
**Summary:** For a section to be marked as ‘completed’ (with a ‘green tick’),  each section has its own set of fields that need to be completed. Currently, it is requiring all fields to be entered before the section set to ‘Completed’ for Personal Details and Address History 

Please see below the list of each required fields in order to make their respective sections ‘Completed’  

Note, please see bug relating to saving changes in the 'Address History': https://github.com/reapit/foundations/issues/572

**Personal Details**
Title
Forename
Surname
Date of Birth
At least one telephone number is required
 
**Address History**
Line 1
Line 3
Postcode
Number of Years at Address
Number of Months at Address
Document type
UPLOADED FILE

It is not a requirement to fill in the ‘Less that 3 years’ section in order to make ‘Address History’ complete

PR #672, opened 2020-03-20
fix: #596 condition to check completed status

leXuanNha:
fixes: #596


packages/elements/src/utils/completed-section/completed-sections.ts
```diff
@@ -33,32 +30,14 @@ export const isCompletedSecondaryID = (identityCheck?: IdentityCheckModel | null
...
     !!contact?.primaryAddress?.line1 &&
-    !!contact?.primaryAddress?.line2 &&
     !!contact?.primaryAddress?.line3 &&
-    !!contact?.primaryAddress?.line4 &&
+    !!contact?.primaryAddress?.postcode &&
     !!contact?.metadata?.primaryAddress?.year &&
     !!contact?.metadata?.primaryAddress?.month &&
     !!contact?.metadata?.primaryAddress?.documentType &&
     !!contact?.metadata?.primaryAddress?.documentImage
-  const isValidSecondaryAddress =
```
tanphamhaiduong: Why do we need to remove this one?

leXuanNha: Because It is not a requirement to fill in the ‘Less that 3 years’ section in order to make ‘Address History’ complete

Issue #568, opened 2020-03-10
Refactor: Refactor auth context for no redux options in scaffolding


tanphamhaiduong:
Context:
Auth flow for no redux options in scaffolding quite complexity

Ouput
Refactor: auth context for no redux options in scaffolding to more simpler and add test

willmcvay:
Tested this locally and throws an error when logging out am afraid, sorry.

<img width="942" alt="Screenshot 2020-03-19 at 10 16 45" src="https://user-images.githubusercontent.com/4630495/77057527-0072f480-69cc-11ea-9df1-66231e1b0ceb.png">

tanphamhaiduong:
@willmcvay could you give me more detail about the step to replicate it?

tanphamhaiduong:
@willmcvay I think there is lack of config, so that is the reason why it show error redirect_uri is not present

willmcvay:
@tanphamhaiduong Sorry but am getting a `refresh-user-session.ts:20 Bad request, failed to refresh session, Error: Both UserPoolId and ClientId are required.` error. I think it's because the UserPoolId needs to be included in the bas config. This is fine to commit as it's just an id not a key or token.

tanphamhaiduong:
@willmcvay you should add UserPoolId and ClientId of new app to config.json before using it.

PR #688, opened 2020-03-20
chore: #629 migrate config.json for scaffolder and ts defitions

tanphamhaiduong:
Changes
- Move scripts ts-definitions to correct folder
- Remove unuse code from fetch definition


packages/config-manager/src/index.js
```diff
@@ -1,191 +1,10 @@
 const AWS = require('aws-sdk')
 const fs = require('fs')
 
-const { generateConfigTsDef } = require('./utils/generate-config-ts-def')
```
willmcvay: We can't delete config manager code am afraid - it's a public NPM package, needs to be kept alive

willmcvay: We also document it as a way of storing secrets

tanphamhaiduong: @willmcvay So could you describe more for me. Because we'll not use reapit-config.json anymore. So ts-config will be scaffold by react scaffolder and in the global.d.ts

tanphamhaiduong: @willmcvay so i'll revert config-manager and just add fetch config for the new one function

tanphamhaiduong:
For the react scaffolder I will retest all again by this issue https://github.com/reapit/foundations/issues/568

Issue #635, opened 2020-03-16
As a ‘Developer’, I want to review my pending revision and have the option to cancel it


HollyJoyPhillips:
Given that I am a ‘Developer’, when I have submitted a revision on my app, I should then have the option to see what revisions are awaiting approval and to also have the option cancel that revision if required 

**Summary:** Currently when a Developer makes a revision to their app, they cannot see what revisions are awaiting approval or have the option to cancel the revision. App revisions can only be ‘Rejected’ or ‘Approved’ by an ‘Admin’. In some instances, a ‘Developer’ may wish to make additional changes or cancel the current changes so we should present what is waiting to be approved and a ‘Cancel Pending Revision’ option. 

**Task:** To display in a modal the revision(s) that have been submitted and add a ‘Cancel Pending Revisions’ button 

- When an app has ‘pendingRevisions’ set to true. We should enable the ‘Pending Revisions’ button (currently disabled) in the Developers Portal
- When clicked, it should open a modal displaying the Apps details and highlighting the changes that have been made using the same diffing functionality as we do in the ‘Admin Portal’ to show what is waiting to be approved. All fields should be locked (as is on the Admin portal (https://dev.marketplace.reapit.cloud/admin/approvals) 
- Add a ‘Cancel Pending Revisions’ button (
- When clicked, present confirmation of cancellation
Modal Title: Please confirm 
Text: Are you sure you wish to cancel any pending revisions for this App?  
Buttons: ‘Yes, proceed’ and ‘Cancel’
- Should post to '/apps/{id}/revisions/{revisionId}/reject' with Developers Email, Name (obtained through the login session) and set rejectionReason as 'Developer Cancelled' 
- When confirmed (Yes, Proceed), present success message
- Success message wording: All pending revisions for this app have been cancelled. You can now use the ‘Edit Detail’ option to make any additional changes as required.
Button ‘Back to App’ should close the success modal 
- The ‘Edit Detail’ button should be present again 

Current UI when a revision is pending approval: 
 ![image](https://user-images.githubusercontent.com/52408607/76775599-79d8d000-679d-11ea-9aac-11394073a7fb.png)

Mockup of new UI: 
![image](https://user-images.githubusercontent.com/52408607/76775617-80ffde00-679d-11ea-84da-982c8bdf473b.png)

When 'Pending Revision' is clicked: 
![image](https://user-images.githubusercontent.com/52408607/76775637-8bba7300-679d-11ea-9948-4512d520c715.png)

[...]

PR #704, opened 2020-03-24
Feat: #635 developers can review and cancel pending revision

nphivu414:


packages/marketplace/src/reducers/revision-detail.ts
```diff
@@ -57,6 +57,8 @@ const revisionDetailReducer = (state: RevisionDetailState = defaultState, action
       loading: false,
       error: false,
       revisionDetailData: action.data,
+      approveFormState: 'PENDING',
```
tanphamhaiduong: Does it same `'PENDING'` like `declineFormState`

packages/marketplace/src/sagas/revisions.ts
```diff
@@ -0,0 +1,46 @@
...
+import { URLS, generateHeader } from '../constants/api'
+import { put, call, takeLatest, all, fork } from '@redux-saga/core/effects'
+import ActionTypes from '../constants/action-types'
+import { Action } from '../types/core'
+import { errorThrownServer } from '../actions/error'
+import errorMessages from '../constants/error-messages'
+import { revisionsReceiveData, revisionsRequestDataFailure, RevisionsRequestParams } from '@/actions/revisions'
+import { logger } from 'logger'
+
+export const fetchAppRevisions = async (params: RevisionsRequestParams) => {
```
tanphamhaiduong: suggest to move this function to `api.ts`

packages/marketplace/src/reducers/revision-detail.ts
nphivu414: Yes, they're using the same values. This is the defaultState:
```
export const defaultState: RevisionDetailState = {
  loading: false,
  error: false,
  revisionDetailData: null,
  approveFormState: 'PENDING',
  declineFormState: 'PENDING',
}
```

packages/marketplace/src/components/ui/app-revision-comparision/app-revision-comparision.tsx
```diff
@@ -0,0 +1,161 @@
...
+      type: revisionMedia?.type || '',
+    }))
+  }
+
+  return app.media.map((currentMedia: MediaModel, index: number) => ({
+    changedMedia: revision.media?.[index]?.uri,
+    currentMedia: currentMedia?.uri,
+    order: currentMedia?.order || 0,
+    type: currentMedia?.type || 'media',
+  }))
```
vuhuucuong: Do we need to use `?` here?
if `revision.media` or `app.media`  is `undefined`, this function will return `undefined`, which not match `DiffMediaModel[]`. Should check manually rather than using opt chain`?`

packages/marketplace/src/components/ui/developer-app-revision-modal/developer-app-revision-modal.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+export const backToAppDetailsModal = (fetchAppDetail, clearAppRevisions, clearAppRevisionDetail, appId) => {
+  return () => {
+    clearAppRevisions()
+    clearAppRevisionDetail()
+    fetchAppDetail(appId)
+  }
+}
+
+export const handleUseEffecttoFetchAppRevisions = (appId, fetchAppRevisions, visible) => {
```
vuhuucuong: should be `handleUseEffectToFetchAppRevisions`

packages/marketplace/src/components/ui/developer-app-revision-modal/developer-app-revision-modal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  return () => {
+    if (appId && visible) {
+      fetchAppRevisions(appId)
+    }
+  }
+}
+
+export const handleUseEffecttoFetchAppRevisionDetail = (appId, appRevisionId, fetchAppRevisionDetail, visible) => {
```
vuhuucuong: should be `handleUseEffectToFetchAppRevisionDetail `

Issue #652, opened 2020-03-18
Uploading a CSV to table should validate data before submitting to server


willmcvay:
Background; 

See #651 We will be processing data in a queue when uploading a CSV of records to a table component.  

Feature;

To support this, the initial upload should trigger the standard validation for that entity, on each item in the CSV. If errors are found, should show a dialogue asking the user if they want to re-upload or process the valid records only. Yes to proceed to POST / mutation on the filtered list, no to junk the CSV.

Should also validate the maximum size of the CSV - this should be a variable configurable by entity as the optimal max size will vary. Use 30 records as an initial default until we know how it performs.

vuhuucuong:
@HollyJoyPhillips FYI, you can QA this in storybook in here:  https://dev.elements.reapit.cloud/?path=/story/spreadsheet--upload-with-validate
In smb, we need to wait for @trankhacvy to integrate it

PR #750, opened 2020-03-27
feat: #652 Upload CSV should validate before submit

vuhuucuong:
fixes: #652 
- Add callback props `afterUploadDataValidated` to run after validating upload data
- Add modal to show after validating upload data
- Upload is no longer replace all Spreadsheet, the logic is now customizable in `afterUploadDataValidated`


packages/elements/src/components/Spreadsheet/handlers.tsx
```diff
@@ -217,19 +220,83 @@ export const handleClickUpload = (ref: React.RefObject<HTMLInputElement>) => ()
...
-}) => {
-  const { target } = event
-  if (target && target.files && target.files[0]) {
-    const file = target.files[0]
-    const result = await parseCsvFile(file)
-    const compatibleData = convertToCompatibleData(result)
-    const dataWithIsValidated = validatedDataGenerate(compatibleData, validate)
-    setData(dataWithIsValidated)
-    return 'validated'
+export const handleOnChangeInput = ({
```
tanphamhaiduong: This function too long, I think you should break down to smaller functions

PR #762, opened 2020-03-30
chore: fix edit release not trigger github action

vuhuucuong:
- We need to capture the `release: edited` event, but it's not being triggered when authenticating with `GITHUB_TOKEN` . Use `GH_PAT` (personal access token) instead. As suggested in here:
https://help.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
- Fix duplicated `_release` like [this](https://github.com/reapit/foundations/releases/tag/elements_v0.5.43) to avoid GitHub action being triggered 2 times


phmngocnghia:
LGTM

`editReleaseNote` is handled by `release-prod.js` script

Issue #725, opened 2020-03-25
As a 'Developer', I want to be able to filter the 'Detailed' page to customise the results


HollyJoyPhillips:
Dependent on: #710

Given that I am a 'Developer', when I am on the 'Detailed' page of the 'Analytics', I should then have the option of filtering the results of the page by Date, Client and App. 

Task: To create a filter bar on the top of the 'Detailed' page to apply filtering to the whole page. 

- Should add 3 filters: 
**Date** to and from (use calendar picker component https://dev.elements.reapit.cloud/?path=/story/datepicker--primary
**Client** (Display a list of ClientID that have installed apps against this developer, using the /installations endpoint. We already use this on the main Analytics page under Client on the 'Installations' table
**App** (display the AppName associated to the developer) 
- by default when landing on this page the date range should be last week (the last Monday to Sunday from today's date), client 'All' and App 'All' 
- Client should be a drop down menu with the option of 'All' 
- App should be a drop down menu with the option of 'All' 
- Should also add default filters 'Yesterday', 'Last Week' and 'Last Month' 
- Yesterday should filter the previous day 
- Last week should filter (the last Monday to Sunday from today's date)
- Last Month should filter against the last calendar month from the current month

See mock up of filter options: 
![image](https://user-images.githubusercontent.com/52408607/77665173-32a9c680-6f77-11ea-8b49-7b7982211b51.png)

See mock up of complete page the filter will apply to: 
![DashBoardDetailedTab](https://user-images.githubusercontent.com/52408607/77665196-405f4c00-6f77-11ea-92dd-49024ae59909.png)

PR #777, opened 2020-04-01
Feat: #725 developer can filter detailed page to customize results

nphivu414:


packages/marketplace/src/components/ui/analytics/detailed/filter-bar/filter-bar.tsx
```diff
@@ -0,0 +1,89 @@
...
+    developerAppIds,
+  }
+}
+
+export const prepareClientIds = installationAppDataArray => {
+  const clientIds = installationAppDataArray
+    .map((installation: InstallationModel) => {
+      return installation.client || ''
+    })
+    .filter((x, i, a) => a.indexOf(x) == i)
```
tanphamhaiduong, line 48: What is x, i a, so I think pls change the name make more sense.

PR #778, opened 2020-04-01
Feature/651 upload table data on worker queue

trankhacvy:


packages/smb/src/hooks/use-upload.ts
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
+import { UploadStateContext, UploadDispatchContext } from '@/context'
+
+export function useUploadState() {
```
willmcvay, line 4: I much prefer this, our hooks are separated by concern 👍

packages/smb/src/utils/worker-upload-helper.ts
```diff
@@ -0,0 +1,114 @@
...
+    })
+      .then(response => response.json())
+      .then((response: ExecutionResult) => {
+        if (response.errors) {
+          resolve({
+            success: false,
+            data: { operationName, query, variables },
+            errors: response.errors.map((error: GraphQLError) => ({ message: error.message })),
+          })
+        } else {
```
willmcvay, line 82: no need for else here if you have resolved at the previous step

packages/smb/src/core/upload-provider.tsx
```diff
@@ -0,0 +1,123 @@
...
+  dispatch(resetState())
+  setCompleted(false)
+}
+
+function UploadProvider({ children }: UploadProviderProps) {
+  const [isCompleted, setCompleted] = React.useState(false)
+  const [state, dispatch] = React.useReducer(reducer, initialState)
+  const { status, totalItem, completedItem, uploadResult } = state
+
+  React.useEffect(() => {
```
tanphamhaiduong, line 97: pls make sure test cover this function

packages/smb/src/types/global.d.ts
```diff
@@ -6,6 +6,7 @@ export type Config = {
   googleAnalyticsKey: string
   cognitoOAuthUrl: string
   cognitoUserPoolId: string
+  accessToken: string
```
tanphamhaiduong, line 9: could we use localstorage instead of add it to config ?

packages/smb/src/core/index.tsx
```diff
@@ -17,6 +17,7 @@ window.reapit = {
     googleAnalyticsKey: '',
     cognitoOAuthUrl: '',
     cognitoUserPoolId: '',
+    accessToken: '',
```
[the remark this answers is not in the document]
trankhacvy: @willmcvay @tanphamhaiduong for now, we cannot access the `window.reapit.config` inside worker thread so I need to pass the access token and graphql backend uri from main thread to work thread. 
I think we should put access token inside the config object because if we save it on localstorage we still need the config object to get the graphql server uri.

packages/smb/src/components/providers/upload-provider/upload-provider.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { resetState } from './actions'
+
+export type UploadProviderProps = { children: JSX.Element }
+
+export const handleCloseModal = (setCompleted, dispatch) => () => {
+  dispatch(resetState())
+  setCompleted(false)
+}
+
+function UploadProvider({ children }: UploadProviderProps) {
```
willmcvay, line 15: This is much tidier 👍

packages/smb/src/components/providers/upload-provider/reducers.ts
```diff
@@ -0,0 +1,61 @@
+import { Cell } from '@reapit/elements'
```
willmcvay, line 1: Can we move to the reducers folder in line with the rest of the app?

packages/smb/src/components/providers/upload-provider/actions.ts
```diff
@@ -0,0 +1,20 @@
+import { Action, UploadResults } from './reducers'
```
willmcvay, line 1: Can we move to the actions folder in line with the rest of the app?

packages/smb/src/types/global.d.ts
willmcvay: So I agree with @tanphamhaiduong here - I see why you have done this now however, configs by definition are static constants whereas the accessToken is stateful so not appropriate. Session storage seems to make most  sense :-)

Issue #739, opened 2020-03-26
The Cloud Apps should be testable in "Desktop Mode"


willmcvay:
- We need to be able to test desktop functionality in standalone mode as a team
- As "desktop mode" relies on a global variable `window.__REAPIT_MARKETPLACE_GLOBALS__ ` we should be able to apply this variable locally for testing without a code change
- Because the apps re-direct to Reapit Connect, and rely on Desktop injecting it at each load, we can't just set this variable manually in the console.
- Suggestion would be to set localstorage as it is persistent, then if a local storage variable is present, execute the set global variable script in the index.ts of each app.
eg

```js
const setMarketplaceGlobals = () => {
    const globals = window.localStorage.getItem('reapit-deskop-globals')
    if (globals) {
        Object.defineProperty(window, '__REAPIT_MARKETPLACE_GLOBALS__', {
            value: JSON.parse(globals),
        })
    }
}
```
Then we can test desktop mode by 

```js
window.localStorage.setItem('reapit-deskop-globals', JSON.stringify({
    cntCode: 'SomeCode'
}))

```
And we can revert to normal mode by

```js
window.localStorage.clear()
```
- Should be exported from elements as a re-usable script (DynamicLinks component), so we can document it for people who also want to test desktop mode.
- Should be applied to all Apps: LTL, AML, Marketplace, Geo Diary, SMB

vuhuucuong:
How to switch mode:
- Open Browser's `Developer Tools` -> `Console` tab
- typing `desktopMode()` to switch to **DESKTOP** mode. If you want to pass data in, use
```
desktopMode({cntCode: 'code'})
```
- typing `webMode()` to switch back to **WEB** mode

PR #784, opened 2020-04-01
feat: #739 The Cloud Apps should be testable in "Desktop Mode"

vuhuucuong:
fixes: #739 
How to switch mode:
- Open Browser's `Developer Tools` -> `Console` tab
- typing `desktopMode()` to switch to **DESKTOP** mode. If you want to pass data in, use
```
desktopMode({cntCode: 'code'})
```
- typing `webMode()` to switch back to **WEB** mode


packages/cognito-auth/src/utils/cognito.test.ts
```diff
@@ -90,6 +90,7 @@ describe('Session utils', () => {
     it('should get a session from the cookie if it exists', () => {
       const stringifiedSession = JSON.stringify({
         refreshToken: mockLoginSession.refreshToken,
+        mode: 'DESKTOP',
```
phmngocnghia, line 93: Would be nice if we have a test case to cover if `marketplaceGlobalObject` returned by `getMarketplaceGlobalsByKey` is falsy: `mode: 'WEB'`

Issue #745, opened 2020-03-27
As a ‘Client’ I want to see a ‘Settings’ icon on my navigation bar


HollyJoyPhillips:
**Summary:** As with the Developers Portal,  we should create a settings icon on the Client Portal which will host the ‘Logout’ option and ‘Customer ID’

Given that I am a ‘Client’ and I am in the Client Portal, when I click ‘Settings’ I should see my Customer ID and have the option to logout

**Task**: To create a ‘Settings’ cog icon/route in the client portal on the nav bar to host the ‘Logout’ CTA and ‘Customer ID’ information
- Add a ‘Settings’ icon using the same cog icon from the developers portal 
- Create a route https://dev.marketplace.reapit.cloud/client/settings 
- Heading ‘Settings’
- Add section ‘Customer ID’ (ClientID)
- Add subtext ‘This is your Customer ID which you will need for use with Private Apps and Web Components. 
- Add the ‘Logout’ CTA here

See Mock up of ‘Settings’ page

![ClientSettings](https://user-images.githubusercontent.com/52408607/77744427-8963de80-7011-11ea-92f6-ad1d091e4adb.jpg)

PR #793, opened 2020-04-03
feat: #745 Add Settings menu for client

trankhacvy:


packages/marketplace/src/components/pages/settings/client-setting.tsx
```diff
@@ -0,0 +1,104 @@
...
+  )
+}
+
+export type StateProps = {
+  customerId: string
+}
+
+export const mapStateToProps = (state: ReduxState): StateProps => {
+  return {
+    customerId: state?.auth?.loginSession?.loginIdentity?.clientId || '',
```
vuhuucuong, line 74: You can safety remove two first `?`. 
```
customerId: state.auth.loginSession?.loginIdentity?.clientId || '',
```
is enough, `?` will vastly reduce test coverage

PR #826, opened 2020-04-06
feat: #728 The search widget back end should return a sub-set of properties data

### LINKED ISSUE #728: The search widget back end should return a sub-set of properties data
- The search widget back end should only return the absolute minimum data it has available from the properties endpoint.
- Task is to map through the response in this file https://github.com/reapit/foundations/blob/master/packages/web-components/src/search-widget/server/api/get-properties.ts and return only the values that are used in the front end code.
- Depends on #622 being merged. You should see from the .svelte files what is being used.
- Should also generate a Pick type, and replace references to PropertyModel in the TS code with this type eg `type SearchWidgetPropertyModel = Pick<PropertyModel, 'val1' | 'val2'>`

vuhuucuong:
fixes #728


packages/web-components/src/search-widget/server/utils/map-minimal-properties.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Map through to return only necessary properties
+ */
+export const mapMinimalProperties = (
+  { _embedded, ...rest }: InputObject,
+  includedEmbeddedProps: string[],
+): OutputObject => {
+  if (!_embedded) {
+    return {
+      ...rest,
+      _embedded,
```
tanphamhaiduong, line 24: I think this `_embedded` will be `undefined` or `null`. So should it be `_embedded: []` ?

vuhuucuong: No, if `_embedded `is undefined, then it should return the original object so that it will be compatible with type from `@reapit/foundations-ts-definitions`

packages/web-components/src/search-widget/types/PickedPropertyModel.d.ts
```diff
@@ -0,0 +1,19 @@
+import { PropertyModel, PagedResultPropertyModel_ } from '@reapit/foundations-ts-definitions'
+
+export type PickedPropertyModel = Pick<
```
willmcvay, line 3: 👍

packages/web-components/src/search-widget/server/api/get-properties.ts
```diff
@@ -4,19 +4,34 @@ import { PagedResultPropertyModel_ } from '@reapit/foundations-ts-definitions'
...
     const url = new URL(`${process.env.PLATFORM_API_BASE_URL}${req.url}`)
-    const refreshResponse = await fetcher<PagedResultPropertyModel_, undefined>({
+    const fullPagedResult = await fetcher<PagedResultPropertyModel_, undefined>({
       url: String(url),
       headers,
     })
 
-    if (refreshResponse) {
+    if (fullPagedResult) {
+      const includedProps = [
```
willmcvay, line 20: Can we move to a constant file so the same list is included in the unit tests - there is a good chance this list will be added to so would be nice if the test can be accurate for free in this case.

packages/web-components/src/search-widget/server/utils/map-minimal-properties.ts
```diff
@@ -0,0 +1,34 @@
...
+type PickedModel = PickedPropertyModel | PickedPropertyImageModel
+type OutputObject = PickedPagedResultPropertyModel_ | PickedPagedResultPropertyImageModel_
+
+/**
+ * Map through to return only necessary properties
+ */
+export const mapMinimalProperties = (
+  { _embedded, ...rest }: InputObject,
+  includedEmbeddedProps: string[],
+): OutputObject => {
```
willmcvay, line 20: Rather than this union type, can mapMinimalProperties not accept a generic eg `const mapMinimalProperties = <T, R>(...` where T is the input and R is the result type?

willmcvay: Not a big deal, just a thought

vuhuucuong: That's great, fixed 👍

PR #829, opened 2020-04-06
chore: #788 fix kind-of vulnerabilities

### LINKED ISSUE #788: Address security vulnerability in kind-of package
Screenshot 2020-04-02 at 13 55 51

See screenshot, looks like a sub-dependency of some dev-dependencies. Lerna and Sass look like the main culprits here.

vuhuucuong:
fixes #788 
Use `kind-of ^6.0.3` as people currently doing in here https://github.com/jonschlinkert/kind-of/pull/31

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
     "**/**/styled-components": "4.4.1",
     "**/**/react-test-renderer": "16.12.0",
     "acorn": "^6.4.1",
-    "minimist": "^1.2.2"
+    "minimist": "^1.2.2",
+    "kind-of": "^6.0.3"
   },
   "dependencies": {
     "@sentry/browser": "^5.11.1",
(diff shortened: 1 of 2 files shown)


mgk:
Just curious, did you just edit `yarn.lock` directly or run `yarn`. On one of my projects when I attempted what you did here running `yarn` after editing `yarn.lock` as you have here, `yarn` spat out a warning and put back the old versions.

vuhuucuong:
@mgk no, you shouldn't edit yarn.lock directly, you should use https://classic.yarnpkg.com/en/docs/selective-version-resolutions/
Basically, It allows us to override the version of the sub-dependency, which means the dependency of our dependency

mgk:
Yeah, that's been my mantra too (not editing yarn.lock). Thanks very much for the pointer.

PR #845, opened 2020-04-08
fix: #837 the react scaffolder extremely slow to compile

### LINKED ISSUE #837: The react app scaffolder is extremely slow to compile
- The scaffolder appears to be working fine now, verified however for dev builds it is extremely slow to first paint and seems to be blocked somewhere in webpack - the main thread is blocked and the browser freezes.
- There is a webpack warning you can see in the dev tools saying that the webpack cache environment is not setup in the webpack.config.js - this would be a good starting place to look.
- Looks like a build issue rather than a code issue but could be wrong.
- FYI best way to approach this is to yarn link in the scaffolder folder then yarn link @reapit/generator-react-app-scaffolder in a target folder, then yo @reapit/react-app-scaffolder and scaffold an app to work on the config. When you have debugged the problem, copy your changes back to the main repo.
- The problem is present in all scaffold configurations and can be replicated after the first build, ie when you run yarn start:dev in the app folder.

vuhuucuong:
fixes #837 
### Problem:
The `createGlobalStyle` function of `styled-component` is run at runtime, which means in browser, and our `index.css` file is ~11k lines, causing the browser to freeze
### Fix:
Changing to use `style-loader`, so that the CSS is processed at compile time

### DIFF
--- a/packages/react-app-scaffolder/app/templates/_package.json
+++ b/packages/react-app-scaffolder/app/templates/_package.json
@@ -14,7 +14,7 @@
     <% if (isFoundation) { %>
     <% if (stylesSolution === 'styledComponents') { %>
     "build:prod": "node '../../scripts/webpack/webpack-styled-components-prod.js'",
-    "start:dev": "webpack-dev-server --hot --progress --color --mode development --config ../../scripts/webpack/webpack.base.dev.js",
+    "start:dev": "webpack-dev-server --hot --progress --color --mode development --config ../../scripts/webpack/webpack.sass.dev.js",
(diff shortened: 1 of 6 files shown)


phmngocnghia:
We shoud add `mini-css-extract-plugin` in production mode to receive the benefit of: CSS build time, and caching stuff 👍

vuhuucuong:
> We shoud add `mini-css-extract-plugin` in production mode to receive the benefit of: CSS build time, and caching stuff 👍

I've tested and it doesn't seem to work with `styled-component`

Issue #791, opened 2020-04-02
Developers should be able to register for the public beta


willmcvay:
- Currently the registration routes are commented out. 
- registration should be re-instated
- Should hide/remove the Password and Confirm password field as not required 
- Need to uncomment registration routes and the registration link on the developer login page
- Need to speak to @cbryanreapit to get the registration flow changed in Congnito to self registration
- Test e2e by registering and confirming a new developer.

PR #874, opened 2020-04-09
feat: #792 terms and conditions acceptance should be confirmed by developer API

### LINKED ISSUE #792: Terms and conditions acceptance should be confirmed by developer API
- Currently a developer accepts the terms and conditions on first login and we set a cookie to confirm
- This flow will change with self registration for the public beta
- On registration, on click of the register button, first display the T&Cs modal - only submit if the user accepts.
- If they don't, return back to the registration page, do not register the user
- On acceptance, set the agreedTerms date time on the Developer POST
- T&Cs should be updated to the attached word document
- Depends on #791 

[Reapit-Foundations Developer Agreement (Final v1 03.04.20).DOCX](https://github.com/reapit/foundations/files/4426785/Reapit-Foundations.Developer.Agreement.Final.v1.03.04.20.DOCX)

phmngocnghia:
Fix #792 #791


packages/marketplace/src/components/pages/register.tsx
```diff
@@ -48,19 +49,50 @@ export interface RegisterFormValues {
...
+  handleSubmit: () => void
+}
+export const handleSubmitButtonOnClick = ({
+  handleOpenModal,
+  handleSubmit,
+  setVisible,
+  validateForm,
+}: HandleSubmitParams) => () => {
+  validateForm().then(errors => {
+    console.log('asdas')
```
willmcvay, line 86: Console left in

Issue #862, opened 2020-04-09
The search widget should have a full screen details page


willmcvay:
**Background context or User story:**
_Currently the search widget renders a results list but most of the property detail is not readable. We want users to see the full details of the property on the click of a button_

**Specification or Acceptance Criteria:**
- A link should be added to each of the properties rendering "View details..."
- Should remove the results list and replace with a property detail page, full width, height to fit content
- Should have a "Back to results..." link at the top
- Should hide the number of results message
- Should share the same fields, classes, helpers and themes as the result item component for now - we will expand this later but for v1, the page is just a full size version of the result item

PR #875, opened 2020-04-10
feat: #862 implement property detail screen

trankhacvy:


packages/web-components/src/search-widget/client/components/property-detail.svelte
```diff
@@ -0,0 +1,190 @@
...
+  $: isSelectedProperty = property.id === (selectedProperty && selectedProperty.id) || ''
+
+  const selectProperty = () => {
+    searchWidgetStore.update(store => ({
+      ...store,
+      selectedProperty: property,
+    }))
+  }
+
+  const handleImageError = source => {
```
willmcvay, line 50: I think we could move into a helper - this is used in the result component too

packages/web-components/src/search-widget/client/components/property-detail.svelte
```diff
@@ -0,0 +1,190 @@
...
+    document.documentElement.scrollTop = 0
+  })
+
+  onDestroy(() => {
+    unsubscribeSearchWidgetStore()
+  })
+</script>
+
+<style>
+  .property-item {
```
willmcvay, line 74: Can we follow the component-name convention for classes eg `property-detail-item` makes conflicts less likely

PR #879, opened 2020-04-13
chore: Add tooltip which display when user hover in tag

trankhacvy:


packages/elements/src/styles/components/dropdown-select.scss
```diff
@@ -6,10 +6,10 @@
   display: inline-block;
   position: relative;
   background-color: #fff;
-  border-color: #dbdbdb;
-  border-radius: 4px;
-  color: #363636;
-  border: 1px solid transparent;
+  //   border-color: #dbdbdb;
```
tanphamhaiduong, line 9: Does it temporary comment?

Issue #864, opened 2020-04-09
The search widget results list should be paginated


willmcvay:
**Background context or User story:**
_Currently the search widget returns a maximum of 8 results - this is the desired page size for now however, as a user, I want to be able to navigate by page the results list_

**Specification or Acceptance Criteria:**

- A standard pagination component should sit at the bottom of the results list with forward, back and page numbers within. 
- Forward should fetch the next 8 results, back should fetch the previous 8 results
- Buttons should be disabled if there are no more results in either direction

**Suggested solution:**

- Should be a home rolled paginator component, no libs
- Should have basic black white and grey theme out the box but be themable with the same hover and accent behaviour as the search buttons. Theme should use base font size.

PR #887, opened 2020-04-13
feat: #864 pagination for search widget

tanphamhaiduong:
feat: #864 pagination for search widget


packages/web-components/src/search-widget/client/index.html
```diff
@@ -20,7 +20,7 @@
       ReapitSearchWidgetComponent &&
         new ReapitSearchWidgetComponent({
           theme: window.theme,
-          apiKey: '',
+          apiKey: 'f321789d-f1e0-47a3-a1a4-455fc3b0ec13',
```
willmcvay, line 23: Please don't merge the API key

packages/web-components/src/search-widget/client/components/pagination.svelte
```diff
@@ -0,0 +1,92 @@
...
+  const unsubscribeSearchWidgetStore = searchWidgetStore.subscribe(store => {
+    totalPage = store.totalPage
+    pageNumber = store.pageNumber
+  })
+
+  onDestroy(() => {
+    unsubscribeSearchWidgetStore()
+  })
+
+  const handlePageChangeClick = (pageNumber) => async() => {
```
willmcvay, line 19: Why not set the current page on the store, then call fetchProperties in the onUpdate hook in the search form - would mean we don't have to copy paste the fetching of properties and images in here and we could re-use the paginator elsewhere

packages/web-components/src/search-widget/client/components/pagination.svelte
```diff
@@ -0,0 +1,92 @@
...
+
+<style>
+  .pagination {
+  display: inline-block;
+}
+
+.pagination a {
+  color: black;
+  float: left;
+  padding: 8px 16px;
```
willmcvay, line 62: Cam we use ems please so will re-scale with the font size when themed eg 0.5em 1em

packages/web-components/src/search-widget/client/components/pagination.svelte
```diff
@@ -0,0 +1,92 @@
...
+  text-decoration: none;
+  transition: background-color .3s;
+}
+
+.pagination a.active {
+  background-color: #4CAF50;
```
willmcvay, line 68: Colors should come from the theme class

tanphamhaiduong:
@willmcvay sorry, this PR WIP

packages/web-components/src/search-widget/client/components/pagination.svelte
tanphamhaiduong: @willmcvay so do you mean https://svelte.dev/docs#afterUpdate and https://svelte.dev/docs#beforeUpdate? So I think cannot use store.update in these hooks, because it'll cause infinity re-render for the component. So do you have another idea?

tanphamhaiduong: @willmcvay fixed as your comment

Issue #830, opened 2020-04-06
As a ‘Developer’, I want to select integrations type for my App to work within Agency Cloud


HollyJoyPhillips:
**Summary**: We will be supporting app integration types to work with Agency Cloud on the submit/edit app forms using a new tag component

Given that I am a ‘Developer’, when I am submitting my app, I should then have the option of searching and selecting single or multiple Agency Cloud Integration types using a tagging component 

**Dependent on ticket:** https://github.com/reapit/foundations/issues/828

**Task:** To create a section on the 'Submit/Edit' app form to support Agency Cloud Integration types for desktop 

- **Heading**: Agency Cloud Integration 
- **Sub text:** To be able to associate your application with an action in Agency Cloud the application will need to be given a desktop type. Please select the type of integration your app requires from the list below. For more information on Desktop Types, please click here (link to https://foundations-documentation.reapit.cloud/api/desktop-api#desktop-types) 
- Should have a dropdown list with the desktop types 
- Endpoint /DesktopIntegrationTypes
- Should use the tagging component to display which options have been selected (see dependent ticket)
- Create heading 'Types Selected' when a tag has been used
- If not tags have been selected, do not display 'Types Selected' heading 
- Should be editable when editing the app
- Should display the ‘description´ when hovering over a selected tag 
- Add text in the hover over, after the description 'More Info.' and link to 'url' field. 
- Location: Between 'App Listing' and 'Authentication Flow' 
- Field to post to /Apps / desktopIntegrationTypeIds
- Platform ticket: https://github.com/reapit/foundations/issues/702

See mockup when first loading the Submit App form:
![AgencyCloudTypes](https://user-images.githubusercontent.com/52408607/78550060-57196480-77fb-11ea-932d-6ae5588af995.jpg)

See mockup when a tag(s) has been selected: 
![AgencyCloudTypesSelected](https://user-images.githubusercontent.com/52408607/78550238-a2cc0e00-77fb-11ea-98b9-dc4072ce25d1.jpg)

See mockup of hover over for description: 
![AgencyCloudTypesInfo](https://user-images.githubusercontent.com/52408607/78551896-6352f100-77fe-11ea-88c7-179088bebd61.jpg)

HollyJoyPhillips:
@trankhacvy, thanks for this, a couple of points:
Please can you hyper link the 'click here' and remove the (link to...) from the text

[...]

PR #928, opened 2020-04-16
chore: #830 Update tag link component and option font size

trankhacvy:


packages/elements/src/styles/components/dropdown-select.scss
```diff
@@ -205,6 +205,7 @@
 
 .rc-select-item-option {
   position: relative;
+  font-size: 14px;
```
tanphamhaiduong, line 208: I think should use rem instead `px`

Issue #863, opened 2020-04-09
The property search widget detail page should have an image lightbox


willmcvay:
**Background context or User story:**
_See #862 given I have a property detail page, I want to be able to see more images of the property in a lightbox_

**Specification or Acceptance Criteria:**
- Should render a clickable list of images below the main image
- Should have a carousel that opens in a lightbox on click of an image
- Use this as a basis for the behaviour: https://www.rightmove.co.uk/property-for-sale/property-77038237.html

**Suggested solution:**
- Should not use any HEAVY libraries - the search widget is only 130kb in prod and we need to stay super lighweight
- If there is no super lightweight carousel / lightbox component please roll our own as a common component in Svelte
- We should consider it as a possible open source package if we do roll our own
- Has to support IE 11 and should be CSS driven as much as possible

PR #937, opened 2020-04-16
feat: #863 - lightbox component for search widget

phmngocnghia:
feat #863


packages/web-components/src/search-widget/client/components/light-box/widget.svelte
```diff
@@ -0,0 +1,180 @@
...
+  }
+</script>
+
+<style>
+  :global(div.light-box-widget-container.light-box-widget-container) {
+    background: #00000045;
+    display: flex;
+    color: white;
+
+    margin-top: -40px;
```
willmcvay, line 122: Please can we use ems for spacing

willmcvay: I completely agree @phmngocnghia, these components are all about keeping it lightweight and fast - nice work 👍

PR #960, opened 2020-04-20
fix: revert web-component common modal

ant066:


packages/web-components/src/common/components/modal.svelte
```diff
@@ -18,79 +17,12 @@
...
-        <h3 class="modal-title" {title}>{title}</h3>
-        <div class="modal-content">
-          <slot />
-        </div>
-      </div>
+      <slot />
     </ClickOutSide>
   </div>
-{/if}
+{/if}
```
tanphamhaiduong, line 28: Pls add more new line to the end

Issue #934, opened 2020-04-16
Create query to for Appointment Types in GEO Diary 


HollyJoyPhillips:
**Summary:** Create Apollo server side query to populate the ‘Appointments Types’ in GEO Diary

**Task:** To write in Apollo Server package

- Query Apollo service to get Appointments Types

Endpoint:**  https://dev.platform.reapit.cloud/configuration/appointmentTypes

PR #962, opened 2020-04-20
feat: #934 - Create GraphQL resolver for PropertyTypes in GraphQL Server

phmngocnghia:
Feat #934


packages/graphql-server/src/constants/api.ts
```diff
@@ -7,6 +7,7 @@ export const URLS = {
   appointments: '/appointments',
   identityChecks: '/identityChecks',
   negotiators: '/negotiators',
+  propertyTypes: '/configuration/appointmentTypes',
```
willmcvay, line 10: So I think I have screwed up here with my ticket - really sorry, I thought the name was PropertyTypes but should be AppointmentTypes. Apologies. 


Also, is there an easy way to expand this endpoint so we can capture all of the configuration options by passing in an extra param eg `/configuration/:type` They all should return the same key value pairs so should be possible right? Would be a really nice enhancement if so :-)

packages/graphql-server/src/constants/api.ts
```diff
@@ -7,7 +7,12 @@ export const URLS = {
   appointments: '/appointments',
   identityChecks: '/identityChecks',
   negotiators: '/negotiators',
+<<<<<<< HEAD
```
willmcvay, line 10: Merge conflicts!

packages/graphql-server/src/resolvers/index.ts
```diff
@@ -5,7 +5,12 @@ import Areas from './areas/resolvers'
 import Offices from './offices/resolvers'
 import Appointments from './appointments/resolvers'
 import Negotiators from './negotiators/resolvers'
+<<<<<<< HEAD
```
willmcvay, line 8: Merge conflicts again, sorry

PR #972, opened 2020-04-20
feat: #939 Web apps should have their own dev and prod cookie identifier

### LINKED ISSUE #939: Web apps should have their own dev and prod cookie identifier
_This story is actually to fix a bug. It is possible to get two cookies for the same app, meaning logout fails because there are two Reapit Session cookies. To solve, each app should have it's own cookie identifier for dev and prod_

- To replicate problem. Log into prod marketplace having first cleared your cookies. Then in the search bar, add `dev.` in front of `marketplace` then  refresh the page. You will have two cookies.
- The cookie identifiers should be unique to dev and prod to avoid this behaviour. 
- Needs to be applied to all web apps that use reapit-connect

**Suggested solution:**
- For login and log our we should rely on the reapit.config.appEnv to determine which cookie to create and to destroy.

vuhuucuong:
fixes #939 
- Web Apps now have three different cookie identifiers `local`,  `developement` , `production`, based on `config.json`
- Some methods in `cognito-auth` now have an additional `appEnv?: string` param to work with cookie for different environments
- Default `refreshSession` state in `redux store` now will be set after fetch `config.json`
- Fix e2e tests login after changing cookie identifiers


packages/geo-diary/src/reducers/auth.ts
```diff
@@ -11,14 +11,17 @@ import {
 import { RefreshParams, LoginSession, getSessionCookie } from '@reapit/cognito-auth'
 import { COOKIE_SESSION_KEY_GEO_DIARY } from '../constants/api'
 import { getMarketplaceGlobalsByKey } from '@reapit/elements'
+import config from '../../config.json'
```
tanphamhaiduong, line 14: This one will cause error when async happen

packages/aml-checklist/src/reducers/auth.ts
```diff
@@ -3,6 +3,7 @@ import { isType } from '@/utils/actions'
 import { authLogin, authLoginFailure, authLoginSuccess, authLogoutSuccess, authSetRefreshSession } from '@/actions/auth'
 import { RefreshParams, LoginSession, getSessionCookie } from '@reapit/cognito-auth'
 import { COOKIE_SESSION_KEY_AML_APP } from '../constants/api'
+import config from '../../config.json'
```
tanphamhaiduong, line 6: I think cannot import config.json directly because when async load happen will cause app crash

packages/geo-diary/src/reducers/auth.ts
vuhuucuong: I think no, because we run `yarn fetch-config` before `yarn build`, and the `config.json` is always available. I cannot use `window.reapit.config.appEnv` because it's `async`

packages/aml-checklist/src/reducers/auth.ts
vuhuucuong: Got your idea, I'm fixing it

PR #978, opened 2020-04-20
fix: adjust access token expiry to 5 minutes

nphivu414:


nphivu414:
Yes @willmcvay , I'll release the new version to NPM after this PR is merged

willmcvay:
@nphivu414 Would you mind waiting on @vuhuucuong 's PR as it would be good if both can get released together as both effect the Cognito Auth service - thanks

willmcvay:
Waiting to release to NPM I mean, not merging!

PR #1092, opened 2020-04-28
fix: #1065 fix loading chunk fail

### LINKED ISSUE #1065: Create CRON to purge S3 buckets based on commit hash

_We get occasional chunk missing errors in dev when we deploy because currently we replace the content of a bucket with the new deployment. If a user has a session, the script will be missing when they navigate. eg of this behaviour https://sentry.io/organizations/reapit-ltd/issues/1623235951/?project=1885603&referrer=slack_

- We should support at least 2 recent versions of each app in an S3 bucket. 
- We should update our webpack scripts so the master HEAD commit hash is prepended to the output script file names. This means we can identify the scripts by release
- We should have a CRON job that runs weekly to purge all S3 buckets of all but the last two releases by most recent commit hash.

phmngocnghia:
Fixes #1065


.github/workflows/purge-unused-assets-cronjbo.yml
```diff
@@ -0,0 +1,45 @@
+name: Purge unused assets from s3
```
willmcvay, line 1: Think the filename has a typo cronjob!

packages/react-app-scaffolder/app/templates/base-is-not-sass/src/scripts/webpack-prod.js
```diff
@@ -8,12 +8,16 @@ const { EnvironmentPlugin } = require('webpack')
 const { PATHS } = require('./constants')
 const path = require('path')
 
+// HEAD~1: Hash appended to files in commit N will be generated by commit N - 1
```
tanphamhaiduong, line 11: I think the better one is HEAD~2, we should support the latest version and latest-1 version. It'll be more safer

PR #1109, opened 2020-04-29
Fix 1052 both select appointment on demo site opens book a valuation

### LINKED ISSUE #1052: Both 'Select Appointment' on Demo site opens Book a Valuation
**Summary:** We have now added 'Book a Viewing' and 'Book a Valuation' on the Demo Site but both options open 'Book a Valuation'

**Relates to ticket:** https://github.com/reapit/foundations/issues/938

The button on the home page (top right hand corner) should be for a Valuation (as it currently is) 
image

However, the button on the property details page should be to 'Book a Viewing'
image

**See video:** https://www.dropbox.com/s/2l1x70wemgj9g8y/Demosite.mp4?dl=0

phmngocnghia:
Fixes #1052


packages/demo-site/scripts/build.js
```diff
@@ -11,7 +11,16 @@ const compileTemplate = async () => {
     const ejsTemplate = await fs.readFileSync(ejsFileDir, { encoding: 'utf8' })
 
     const template = ejs.compile(ejsTemplate)
-    const html = template({ API_KEY: config.API_KEY, CDN_URL: config.CDN_URL })
+
+    let CDN_URL = config.CDN_URL
+    const env = process.env.NODE_ENV
+
+    if (env !== 'DEV') {
+      CDN_URL = 'http://localhost:5000'
```
willmcvay, line 19: Sorry, I am a bit confused by this - why are we serving CDN files from localhost in prod?

packages/web-components/package.json
```diff
@@ -45,7 +45,8 @@
   },
   "scripts": {
     "build:prod": "node ./src/scripts/build-prod.js",
-    "start:prod": "sirv public",
+    "start:prod": "serve public/dist -p 5000",
```
willmcvay, line 48: Sorry, am not following what this is for. Can you explain?

packages/demo-site/scripts/build.js
phmngocnghia: It's a typos

packages/web-components/package.json
phmngocnghia: @willmcvay This will serve all bundle asset to localhost:5000, there is a command named `yarn start:dev` the the demo-site package which will:

- Bundle web-components and serve it to localhost:5000
- Bundle demo-site with CDN_URL = localhost:5000
- Run the serverless package
- It's allow us to test the demo-site against local web-components package

I haven't figured out the way to make it "live reload".

willmcvay: Okay, so I get it @phmngocnghia thanks. So maybe start:prod is confusing me here - maybe start:dev would make more sense. start:prod was implying we would launch in production. Will leave with you though 👍

phmngocnghia: @willmcvay I will change it to serve:prod as it implies we are serving prod static files, and not for launch

Issue #1122, opened 2020-04-30
As a client, I should redirect to the browse apps page if I have no installations


willmcvay:
**Background context or User story:**

_Currently when loading the client side of marketplace the default is to land on the installed apps page. If there are no installed apps returned by the initial API call, redirect to the browse apps page_

**Specification or Acceptance Criteria:**

- This behaviour should only apply on the initial login so I don't end up in a redirect loop when navigating from the sidebar menu.

PR #1133, opened 2020-05-04
feat: #1122 redirect to the browse apps page if I have no installations

ant066:


packages/marketplace/src/components/pages/installed-apps.tsx
```diff
@@ -21,6 +23,12 @@ export const handleOnChange = history => (page: number) => history.push(`${route
 
 export const handleOnCardClick = (app: AppSummaryModel) => handleLaunchApp(app)
 
+export const isFromLoginPage = () => {
+  const isFromLogin = getCookieString(COOKIE_CLIENT_IS_LOAD_INSTALLED_APP_FROM_LOGIN)
+  setCookieString(COOKIE_CLIENT_IS_LOAD_INSTALLED_APP_FROM_LOGIN, '')
```
willmcvay, line 28: Not sure setting another cookie is going to be the best solution here. There is a more redux-friendly way of checking on previous route that seems more appropriate here. The below accepted answer is what I would suggest:

https://stackoverflow.com/questions/49982230/how-to-get-previous-location-with-react-router

ant066: I think when logged at https://dev.connect.reapit.cloud and then redirect to the marketplace, the redux store was cleared, so we can not get the previous location.

willmcvay: Well in that case you know that you re-directed from login or hit the route directly. In which case, if no installed apps > redirect. If has a previous route > don't redirect. Or am I missing something?

ant066: > Well in that case you know that you re-directed from login or hit the route directly. In which case, if no installed apps > redirect. If has a previous route > don't redirect. Or am I missing something?
If use this solution, It also redirects to "browse app" page when you access 
"installed apps" by direct link, not login page only.

willmcvay: Yes, I agree. That is what we are trying to achieve. TBH accessing the installed apps page when I don't have any installed apps by direct link is basically an edge case!

tanphamhaiduong:
This PR will be closed right?

PR #1134, opened 2020-05-04
fix #1065: re-instating purge unused s3 assets cronjob

### LINKED ISSUE #1065: Create CRON to purge S3 buckets based on commit hash

phmngocnghia:
- Fixes #1065 
- Remove unused service worker loader, runner installed in the past: (workbox-plugin, CRA)


packages/geo-diary/src/core/__tests__/app.tsx
```diff
@@ -6,6 +6,23 @@ import { render, unmountComponentAtNode } from 'react-dom'
 jest.mock('../router')
 
 describe('App', () => {
+  beforeAll(() => {
+    // @ts-ignore
+    delete global.window.location
+    // @ts-ignore
+    global.window = Object.create(window)
+    // @ts-ignore
+    global.window.location = {
```
willmcvay, line 15: Can you do `jest.spyOn(global, 'window', () => ({ //mock here })`

phmngocnghia: The `window.location` in the test env don't have reload function. Seems like we can't mock not exist function using jest

PR #1136, opened 2020-05-04
chore: add sentry cli to push source map to sentry.io

tanphamhaiduong:
chore: add sentry cli to push source map to sentry.io


scripts/webpack/webpack.base.prod.js
```diff
@@ -3,26 +3,23 @@ const FaviconsWebpackPlugin = require('favicons-webpack-plugin')
...
 const { GenerateSW } = require('workbox-webpack-plugin')
-const HashedModuleIdsPlugin = require('webpack').HashedModuleIdsPlugin
-const { EnvironmentPlugin } = require('webpack')
-const { execSync } = require('child_process')
-// const SentryWebpackPlugin = require('@sentry/webpack-plugin')
-const { PATHS } = require('./constants')
-const { getVersionTag } = require('../release/utils')
-const hashFiles = require('../utils/hash-files')
 const path = require('path')
+const SentryWebpackPlugin = require('@sentry/webpack-plugin')
```
vuhuucuong, line 7: So we're adding sentry source map to `webpack.base.prod.js`, does it mean that when we `yarn:build` in local machine, source map is still pushed to sentry?
Can we only do this when `CI = true` ?https://github.com/reapit/foundations/blob/master/.github/workflows/release-develop.yml#L16

tanphamhaiduong: @vuhuucuong that's my concern, so let me find the way to remove build from PRs, and only accept in release develop and production.

willmcvay: Good spot @vuhuucuong

scripts/webpack/webpack.base.prod.js
```diff
@@ -184,3 +176,20 @@ module.exports = {
     modules: false,
   },
 }
+
+if (process.env.IS_RELEASE) {
```
willmcvay, line 180: 👍

Issue #1086, opened 2020-04-27
Property Details Page - Demo Site 


HollyJoyPhillips:
**Summary:** A few updates required to the Property Details page on the Demo Site 
- When you click on 'View Details' from the property results, the Property Details page should be on its own route using the property ID as the URL. For example: 
If the property ID ="BED130271" the URL should be 'demo-site.reapit.cloud/BED130271'
- Remove the ‘Search option from the top’ should just have a back to results link
- The map feature does not work, if you click in the 'Pin' it looks as though it is sliding but nothing happens. For now, please remove the map slider on the Property Details page 
- For some reason, if you move your mouse over any of this area, it looks clickable but is not: 
![image](https://user-images.githubusercontent.com/52408607/80367303-c296a500-8882-11ea-93d6-afc64a1e304a.png)

HollyJoyPhillips:
@trankhacvy  Hi Vy, 

When I click on 'View Property' for any property, at first i get 'Property not found' 
![image](https://user-images.githubusercontent.com/52408607/81155710-256ff680-8f7d-11ea-8286-16a981937aae.png)


If I then click on 'Back to results' and click on another property, the URL seems to change but the page remains on the property results page: 
![image](https://user-images.githubusercontent.com/52408607/81155806-3caee400-8f7d-11ea-91a7-114bfe54ca7c.png)

willmcvay:
@trankhacvy Really sorry but this issue is going to need a re-think. The ajax routing won't work for us when embedded on someone else's page. It has just broken my site-builder code for example.

- We need the routing to work by re-directing to a new html page eg from index.html > property.html then load a new standalone widget. 
- New widget to be called PropertyDetailWidget
- Will be a significant refactor but most of the code can be re-used am sure.

Again, sorry for having to change the requirement on this.

PR #1149, opened 2020-05-04
feat: #1086 Property detail page

trankhacvy:


packages/web-components/src/search-widget/server/api/get-property.ts
```diff
@@ -0,0 +1,30 @@
...
+  try {
+    const headers = await getServerHeaders(req, PACKAGE_SUFFIXES.SEARCH_WIDGET)
+    const url = new URL(`${process.env.PLATFORM_API_BASE_URL}${req.url}`)
+    const result = await fetcher<PickedPropertyModel, undefined>({
+      url: String(url),
+      headers,
+    })
+
+    if (result) {
+      const minimalResult = pickBy(result, (_, key) => INCLUDED_PROPS.GET_PROPERTIES.includes(key))
```
tanphamhaiduong, line 20: I think you should move this out to new function to reuse like `removeNoNeedProperties`

PR #1170, opened 2020-05-07
chore: pre-commit secrets scanner

vuhuucuong:
Add pre-commit hook to scan for secrets before commit
### FLOW:
1. GET AWS secrets from environment variables and `~/.aws/credentials`
2. Merge those secrets into one file (temporary file)
3. Register AWS provider for git-secrets using the file in step 2
4. Scan for secrets in staged files using `git secrets --scan`
5. Remove temporary file in step 2

### DIFF
--- a/scripts/utils/git-secrets.js
+++ b/scripts/utils/git-secrets.js
@@ -36,7 +36,7 @@ const removeTempCredFile = () => {
 
 /**
  * FLOW:
- * 1. GET AWS secrets from PATH and ~/.aws/credentials
+ * 1. GET AWS secrets from environment variables and ~/.aws/credentials
  * 2. Merge those secrets into one file (temporary file)
  * 3. Register AWS provider for git-secrets using the file in step 2
  * 4. Scan for secrets in staged files using git secrets --scan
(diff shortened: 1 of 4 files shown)


scripts/utils/git-secrets.js
```diff
@@ -0,0 +1,74 @@
...
+    if (stderrAddProvider) throw stderrAddProvider
+    console.log('Added AWS provider\n', stdoutAddProvider)
+
+    // scan for secrets
+    const { error: errorScan, stdout: stdoutScan, stderr: stderrScan } = shell.exec('git secrets --scan --cached', {
+      stdio: 'inherit',
+    })
+    if (errorScan) throw errorScan
+    if (stderrScan) throw stderrScan
+    console.log('No secrets found in staged files!\n', stdoutScan)
```
vuhuucuong, line 65: I tried to chain these two commands using &&, but it doesn't work as expected, need to separate into two commands

scripts/utils/git-secrets.js
```diff
@@ -0,0 +1,74 @@
...
+    const { error: errorScan, stdout: stdoutScan, stderr: stderrScan } = shell.exec('git secrets --scan --cached', {
+      stdio: 'inherit',
+    })
+    if (errorScan) throw errorScan
+    if (stderrScan) throw stderrScan
+    console.log('No secret found in staged files!\n', stdoutScan)
+    removeTempCredFile()
+  } catch (err) {
+    console.error(err)
+    removeTempCredFile()
```
phmngocnghia, line 69: I thinks this part could result unhandled error if `unlinksync` is fail to execute and throws error

LGTM. Nice 👍

vuhuucuong:
> For me i don't use aws credential file, only use bash profile, are there any way to check it ?

@tanphamhaiduong I've already checked both authentication mechanisms.

If you use bash profile, it means you're exporting `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` as environment variables.
I covered it in here https://github.com/reapit/foundations/pull/1170/files#diff-1d82deb06aec9c6a91d1e347cd934da5R11-R27

Just updated comment on that function to avoid confusing

scripts/utils/git-secrets.js
vuhuucuong: Yes if it throws an error, then just let it does, the script will auto exit with an error code.
If all things go well, then no error is returned.

PR #1181, opened 2020-05-11
fix: #1160 HTML Description fields fixes

### LINKED ISSUE #1160: HTML Description field
**Summary:** As we have introduced the ability to edit the Description with HTML, we need to address the following issues 

**Relates to ticket:** https://github.com/reapit/foundations/issues/1123

- Can we make the box the same size as the Summary: 
image

- Adding an external link seems to include the code ‘&nbsp:’	
- The bullet point seems to be out of the green area and not aligned in the revisions modal 
image

 
- Or after saving: 

 
image

- The external link does not work on the modal, when you click it nothing happens: 
image

 
- The link should also be the Reapit blue (#0061a8) 
- We need to be able to view the link in the Admin Portal when a revision is made, might need to be a separate ticket? 
- I can’t edit the external link when editing the app, I would have thought highlighting the wording and clicking on the ‘link’ icon would give me the option to edit the existing link 
 
image

- The line break/paragraph tag doesn’t seem to work at all. 
 
image

vuhuucuong:
fixes #1160


packages/elements/src/styles/vendor/pell.scss
```diff
@@ -14,6 +14,7 @@ $pell-content-padding: 10px !default;
   border-radius: 0;
   overflow: hidden;
   color: $black;
+  height: 180px;
```
tanphamhaiduong, line 17: could we use `rem` instead of `px` ?

packages/elements/src/styles/vendor/pell.scss
```diff
@@ -51,6 +63,16 @@ $pell-content-padding: 10px !default;
     font-weight: bold;
   }
 
+  h6 {
+    display: block;
+    font-size: 1em;
```
tanphamhaiduong, line 68: could we use `rem` instead of `em` ?

packages/elements/src/styles/vendor/pell.scss
```diff
@@ -51,6 +63,16 @@ $pell-content-padding: 10px !default;
...
+    margin-block-start: 0.83em;
```
tanphamhaiduong, line 69: could we use `rem` instead of `em` ?

packages/elements/src/styles/vendor/pell.scss
vuhuucuong: No, because the height must be equal with the existing summary field which has `180px` as height

PR #1186, opened 2020-05-11
feat: #1036 create geo-diary-v2 pipeline

### LINKED ISSUE #1036: Create package and pipeline for Geo Diary V2

_Subject to tickets in progress at time of writing, we support the endpoints necessary to Refactor Geo Diary using Apollo Server rather than Redux. Aim is to showcase GraphQL in prod and to demonstrate reduction in codebase plus better offline support_

- Should clone existing Geo Diary into new repo
- Should have own pipeline so we can QA separately initially
- Pipeline should be dev only

tanphamhaiduong:
resolve #1036 

Changes
- Add geodiary v2 package
- Add pipeline
- Add graphql client and test
- deploy cloudfront


packages/geo-diary-v2/README.md
```diff
@@ -0,0 +1,12 @@
+# geo-diary-v2
+
+geo-diary-v2 for Reapit PAAS platform. Initial scaffold from [React App Boilerplate](https://github.com/reapit/react-app)
+
+## Read on:
+
+- [Getting Started](./src/docs/GETTING_STARTED.md)
```
willmcvay, line 7: Don't think these docs go anywhere now. We should remove from the scaffold at some point...

packages/geo-diary-v2/src/scripts/webpack-dev.js
```diff
@@ -0,0 +1,152 @@
+const path = require('path')
```
willmcvay, line 1: Think the scripts can be removed as the package.json references the root scripts folder. Again, something we should remove from the internal scaffold

Issue #1248, opened 2020-05-13
The submit app page should not be a top level menu item


willmcvay:
**Background context or User story:**

_Currently the submit app page is accessible from the main sidebar menu. That menu is too large now so we should move to be a submit app button in the top right hand side of the "My Apps Page"_

**Specification or Acceptance Criteria:**

- Should have the text "Create new app"
- Should be level with the "My Apps" heading
- Should stack below the heading in mobile
- Routing to and from the page should remain the same

PR #1253, opened 2020-05-14
feat: #1248 Remove submit app menu item. Create new app button in my …

trankhacvy:
…apps page

### DIFF
--- a/packages/marketplace/src/components/ui/app-list.tsx
+++ b/packages/marketplace/src/components/ui/app-list.tsx
@@ -15,7 +15,10 @@ import {
   GridFourColItem,
   Helper,
   infoText,
+  Button,
 } from '@reapit/elements'
+import { history } from '@/core/router'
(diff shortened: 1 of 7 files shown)


packages/marketplace/src/components/ui/app-list.tsx
```diff
@@ -26,6 +29,26 @@ export type AppListProps = {
...
   numOfColumn?: number
+  hasSubmitButton?: boolean
+}
+
+export const renderHeader = ({ hasSubmitButton, title }: { hasSubmitButton: boolean; title: string | undefined }) => {
+  const goToNewAppPage = () => {
+    history.push(Routes.SUBMIT_APP)
+  }
+
+  if (!hasSubmitButton) {
```
tanphamhaiduong, line 40: If !title return null

If !hasSubmitButton && title 
Return h3

Return <div>....

PR #1264, opened 2020-05-15
fix: change from cp to S3 instead of sync

tanphamhaiduong:
Changes
- Add delete from remote bucket repo
- Change from sync to cp to s3 cli


tanphamhaiduong:
I’m still téting it

Issue #1173, opened 2020-05-08
Refactor: STEP 1 Move all API call from marketplace to services folder


tanphamhaiduong:
**Background context or User story:**
- The API call of the service right now in the sagas generator function.
![image](https://user-images.githubusercontent.com/22660833/81368221-6bfa5800-9119-11ea-963b-928d7bf120b9.png)
- For reusable we need to move to `/src/services/<<service_name>>.ts
- What is <<service_name>>? We'll use the name of the API to use as file name.
![image](https://user-images.githubusercontent.com/22660833/81368319-ae239980-9119-11ea-92ad-79671978fd1e.png)

**Specification or Acceptance Criteria:**
- Move all fetcher to call API to `/src/services/` folder and name as service name like the image above
- Add params for fetcher function as query params in swagger to easy extendable in the future if we need to fetch with other params.
![image](https://user-images.githubusercontent.com/22660833/81368376-d14e4900-9119-11ea-9d3f-0e9fc04796be.png)

**Suggested solution:**
- Move to `/src/services/` folder and name as the service name
- Add params like swagger
![image](https://user-images.githubusercontent.com/22660833/81368490-183c3e80-911a-11ea-9f49-eee62a537b78.png)
- Move URLs constants to sagas folder because it only uses in sagas
![image](https://user-images.githubusercontent.com/22660833/81368615-64877e80-911a-11ea-8bbb-9a66275600fe.png)
- Modify sagas call the API to get data and ref to services folder to use and get data via services functions

PR #1265, opened 2020-05-15
refactor: #1173 Step 1 - Move API calls to services folders

vuhuucuong:
fixes: #1173
### What's changed:
- Move all `fetcher` calls to `/services` following swagger params & return types. (all params are flattened)
- Move all related typing defs params to `/services` (should use this when refactoring `action` & `saga` in the next steps for consistency)
- Replace all API calls in sagas to use the above service
- Move all related functions, constants to `/services` (`constants.ts` & `utils.ts`)
- Fix a few typing errors in saga following the above service params (this should be fixed all when we refactoring saga)


packages/marketplace/src/sagas/admin-stats.ts
```diff
@@ -28,15 +27,21 @@ export const adminStatsDataFetch = function*({ data }) {
...
-      url: `${url}?${setQueryParams({
-        pageSize: GET_ALL_PAGE_SIZE,
-        ...queryParams,
-      })}`,
-      api: window.reapit.config.marketplaceApiUrl,
-      method: 'GET',
-      headers: generateHeader(window.reapit.config.marketplaceApiKey),
-    })
+    const serviceToCall =
+      area === 'APPS'
```
tanphamhaiduong, line 31: I think should move to switch case or use hash map to get the condition instead of if else too much
https://medium.com/front-end-weekly/switch-case-if-else-or-a-lookup-map-a-study-case-de1c801d944

packages/marketplace/src/sagas/client.ts
```diff
@@ -29,36 +29,19 @@ export const clientDataFetch = function*({ data }) {
...
-          IsFeatured: isFilteringForDirectApiApps ? undefined : false,
-          IsDirectApi: isFilteringForDirectApiApps ? true : undefined,
-        })}`,
-        api: window.reapit.config.marketplaceApiUrl,
-        method: 'GET',
-        headers: generateHeader(window.reapit.config.marketplaceApiKey),
+      call(fetchAppsList, {
+        clientId,
+        category: isFilteringForDirectApiApps ? undefined : category,
+        [searchBy]: search,
```
tanphamhaiduong, line 35: could we change to `searchBy` ?

packages/marketplace/src/sagas/client.ts
```diff
@@ -29,36 +29,19 @@ export const clientDataFetch = function*({ data }) {
...
-      currentCategories.length > 1
-        ? currentCategories
-        : call(fetcher, {
-            url: `${URLS.categories}`,
-            method: 'GET',
-            api: window.reapit.config.marketplaceApiUrl,
-            headers: generateHeader(window.reapit.config.marketplaceApiKey),
-          }),
+        : call(fetchAppsList, { clientId, pageNumber: 1, pageSize: FEATURED_APPS, isFeatured: true }),
+      currentCategories.length > 1 ? currentCategories : call(fetchCategoriesList, {}),
```
tanphamhaiduong, line 44: Pls avoid magic number `1`

packages/marketplace/src/services/apps.ts
```diff
@@ -1,27 +1,256 @@
...
+  developerName?: string
+  companyName?: string
+  isFeatured?: boolean
+  isDirectApi?: boolean
+  onlyInstalled?: boolean
+  registeredFrom?: string
+  registeredTo?: string
+}
+
+export interface FetchAppByIdParams extends FetchByIdCommonParams {
```
tanphamhaiduong, line 34: So just curious why below is type but this one is interface
So you can declare type extends like this `export type FetchAppByIdParams  = FetchByIdCommonParams & { clientId?: string }`

packages/marketplace/src/services/apps.ts
```diff
@@ -1,27 +1,256 @@
...
+
+export type CreateAppParams = CreateAppModel
+
+export type DeleteAppByIdParams = FetchByIdCommonParams
+
+export type FeatureAppByIdParams = FetchByIdCommonParams
+
+export type UnfeatureAppByIdParams = FetchByIdCommonParams
+
+export interface FetchAppRevisionsListParams extends FetchListCommonParams {
```
tanphamhaiduong, line 46: same as above

packages/marketplace/src/services/developers.ts
```diff
@@ -0,0 +1,89 @@
...
+  PagedResultDeveloperModel_,
+  DeveloperModel,
+} from '@reapit/foundations-ts-definitions'
+import { fetcher, setQueryParams } from '@reapit/elements'
+import { URLS } from './constants'
+import { generateHeader } from './utils'
+import { logger } from 'logger'
+import { FetchListCommonParams, FetchByIdCommonParams } from './types'
+
+export interface FetchDevelopersListParams extends FetchListCommonParams {
```
tanphamhaiduong, line 13: same as above

packages/marketplace/src/sagas/admin-stats.ts
tanphamhaiduong:
```
const services = {
DEVELOPERS: fetchDevelopersList,
INSTALLATIONS: fetchInstallationsList,
APPS: fetchAppsList,
}
const response = yield call(serviceToCall[area], { pageSize: GET_ALL_PAGE_SIZE, ...queryParams })
```

packages/marketplace/src/sagas/client.ts
vuhuucuong: Sorry what do you mean?
Here, `searchBy` is key, and `search` is value

tanphamhaiduong: is searchBy variables ?

vuhuucuong: yes

vuhuucuong: `[searchBy]: search` can be translated into `appName: 'app'` or `developerId: 'id1'`, etc...

tanphamhaiduong: ah it's ok

Issue #101, opened 2020-01-27
As a "developer", I should be directed to information if I try to access the marketplace without a client Id


HollyJoyPhillips:
Given that I am a ‘Developer’, when I want to either login as a ‘Client’, I need to be sponsored by a client with a valid RPS desktop license. Currently, we throw a 'developer id does not exist in state error' and show a 404 message.

**Task**: 

- Do not throw or log the error in Sentry anymore. Instead, show a modal on the current page saying "You do not currently have an Agency Cloud User License so you can't visit the Reapit Marketplace, if you would like more information on Agency Cloud please visit [our public site](https://www.reapit.com/agency-cloud/). Alternatively, if you are a developer on the Reapit Platfom, one of your clients may be willing to sponsor with a paid user license."
- The footer should contain three buttons, "Home Page, Developers, and Logout" redirecting to the respective routes.

nphivu414:
Hi @HollyJoyPhillips ,
The Home Page button will be removed from the modal as what we've discussed because it doesn't make sense here.

Issue #99, opened 2020-01-27
As a ‘Client’, I want to use my existing email address to also register as a ‘Developer’ so that I can use the same email address


HollyJoyPhillips:
Given that I am a ‘Client’, when I want to either login as a ‘Developer’, I need to register. Currently, we throw a 'client id does not exist in state error' and show a 404 message.

**Task**: 

- Do not throw or log the error in Sentry anymore. Instead, show a modal on the current page saying "You are not currently registered as a Reapit Developer, if you would like to register please visit [our registration page](https://dev.marketplace.reapit.com/register)".
- The footer should contain three buttons, "Register, Marketplace, and Logout" redirecting to the respective routes.

tanphamhaiduong:
Assign to @nphivu414 because it relate to his ticket

PR #1270, opened 2020-05-18
feat: #101 Developers should be directed to information if trying to access the marketplace without a client Id

nphivu414:
This PR is related to 2 tickets [#101](https://github.com/reapit/foundations/issues/101) and [#99](https://github.com/reapit/foundations/issues/99)

We will change the behavior when a user login to client portal with a developer account or login to developer portal with a client account. We don't throw or log the error in Sentry anymore. Instead, show a modal to redirect users to the respective routes.


packages/marketplace/src/core/private-route.tsx
```diff
@@ -1,36 +1,76 @@
...
+    return false
+  }
+  const { clientId, developerId } = loginIdentity
+  if (!clientId && !developerId) {
+    return true
+  }
+  return false
+}
+
+export const handleChangeLoginType = (loginIdentity, loginType, allow, dispatch) => {
```
willmcvay, line 29: Can you add definitions please, and to the handleRedirect function

willmcvay:
Looks good, just the small things above but the logic looks sound and clean solution 👍

packages/marketplace/src/components/pages/authentication/__test__/authentication.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  renderClientModal,
+  onLogoutButtonClick,
+  onMarketplaceButtonClick,
+  onRegisterButtonClick,
+  onDevelopersButtonClick,
+} from '../authentication'
+import { authLogout } from '@/actions/auth'
+
+describe('Authentication', () => {
+  const history = {
```
tanphamhaiduong, line 19: I think there is mockRouter on the project, could you find it?

packages/marketplace/src/components/pages/authentication/authentication.tsx
```diff
@@ -0,0 +1,105 @@
...
+      footerItems={
+        <>
+          <Button onClick={onRegisterButtonClick(history)}>Register</Button>
+          <Button onClick={onMarketplaceButtonClick(history)}>Marketplace</Button>
+          <Button onClick={onLogoutButtonClick(dispatch)}>Logout</Button>
+        </>
+      }
+    >
+      <p>
+        {'You are not currently registered as a Reapit Developer, if you would like to register please visit '}{' '}
```
tanphamhaiduong, line 84: I think can remove {' '}
```
You are not currently registered as a Reapit Developer, if you would like to register please visit&nbsp;
```

packages/marketplace/src/constants/routes.ts
```diff
@@ -1,4 +1,6 @@
 const Routes = {
+  Authentication: '/authentication',
+  AuthenticationLoginType: '/authentication/:loginType',
   CLIENT: '/client/apps',
   CLIENT_APP_DETAIL: '/client/apps/:appid',
```
vuhuucuong, line 5: should be uppercase following convention

packages/marketplace/src/core/private-route.tsx
```diff
@@ -1,36 +1,82 @@
...
+}
+
+export const handleRedirectToAuthenticationPage = (loginIdentity, allow) => {
+  return () => {
+    if (!loginIdentity) {
+      return
+    }
+    const { clientId, developerId } = loginIdentity
+    if ((allow === 'CLIENT' && !clientId) || (allow === 'DEVELOPER' && !developerId)) {
+      window.location.href = `${Routes.Authentication}/${allow}`
```
vuhuucuong, line 56: Can we use [useHistory](https://reacttraining.com/react-router/web/api/Hooks/usehistory) from react-router instead?

packages/marketplace/src/constants/routes.ts
willmcvay: I missed this, yes, should be uppercase

PR #1273, opened 2020-05-18
feat: #1221 add cloud alert for production env web apps

### LINKED ISSUE #1221: Cypress test for production environment by cronjob and alert to slack if error
- We should have an alert if the website has a problem

- We'll have cronjob to run any time in the day and check. If the website is down, it'll push an alert to the slack channel

**Suggested solution:**
- We'll have new repo and write E2E test with cypress E2E to check and compare the version like this link https://docs.cypress.io/guides/tooling/visual-testing.html#Functional-vs-visual-testing
- It'll be run by a cronjob every day 1 hour run 1 time
- For serverless backend, we'll have /ping routes and return service health. If it going down push an alert to slack.
- All test only run for production.

tanphamhaiduong:
resolve #1221 
Changes
- Add cypress request cronjob for apps each 30 mins
- Add push notification to cloud-sentry channel if failed
- Clean up un-use cli

### DIFF
--- a/packages/cloud-alert/cypress/support/commands.ts
+++ b/packages/cloud-alert/cypress/support/commands.ts
@@ -1,0 +1,25 @@
+// ***********************************************
+// This example commands.js shows you how to
+// create various custom commands and overwrite
+// existing commands.
+//
+// For more comprehensive examples of custom
+// commands please read more here:
+// https://on.cypress.io/custom-commands
+// ***********************************************
+//
+//
+// -- This is a parent command --
+// Cypress.Commands.add("login", (email, password) => { ... })
+//
+//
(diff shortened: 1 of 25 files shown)


packages/cloud-alert/cypress/support/commands.ts
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
vuhuucuong, line 1: Can we delete this file?

tanphamhaiduong: @vuhuucuong the file come from the template. I think it would be nice if we let there for full structure of cypress project

PR #1449, opened 2020-05-29
fix: #1381 hotfix on admin routing

willmcvay:
# Pull request checklist

fixes: #1381 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- Hotfixes a routing regression in Admin area

Issue Number: #

## What is the new behavior?

- Admin routing working as expected

- [ ] Yes
- [x] No

### DIFF
--- a/packages/marketplace/src/core/private-route.tsx
+++ b/packages/marketplace/src/core/private-route.tsx
@@ -18,12 +18,15 @@ export interface PrivateRouteProps {
   fetcher?: boolean
 }
 
-export const isNotAllowedToAccess = (loginIdentity?: LoginIdentity) => {
+export const isNotAllowedToAccess = (allow: LoginType | LoginType[], loginIdentity?: LoginIdentity) => {
   if (!loginIdentity) {
     return false
   }
-  const { clientId, developerId } = loginIdentity
-  if (!clientId && !developerId) {
+  const { clientId, developerId, adminId } = loginIdentity
+  const isAdminProtected = allow === 'ADMIN' && !adminId
+  const isNotClientOrDeveloper = !clientId && !developerId
+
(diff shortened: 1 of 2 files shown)


packages/marketplace/src/core/private-route.tsx
```diff
@@ -40,14 +43,18 @@ export const handleChangeLoginType = (
...
       return
     }
-    if (loginType === 'CLIENT' && allow === 'DEVELOPER' && loginIdentity.developerId) {
-      dispatch(authChangeLoginType('DEVELOPER'))
+    if (loginType !== 'ADMIN' && allow === 'ADMIN' && loginIdentity.adminId) {
+      dispatch(authChangeLoginType('ADMIN'))
       return
     }
-    if (loginType === 'DEVELOPER' && allow === 'CLIENT' && loginIdentity.clientId) {
+    if (loginType !== 'CLIENT' && allow === 'CLIENT' && loginIdentity.clientId) {
```
tanphamhaiduong, line 50: I think @nphivu414 clear about this one and can answer you what are happening here. This is not the solution for authenticate admin, client, and developer right now. This change will cause another bug

nphivu414: Hi @tanphamhaiduong ,
The new logic here is still correct, it also covers the case when users change URL from client/dev portal to admin portal. The only problem is that this logic looks a bit unclear and it's hard for other members to understand what's going on and why we need to do these things.

tanphamhaiduong: @nphivu414 Thanks for information. @willmcvay  I think this place should change to look more cleaner. Because it's very hard to understand what are happening in here

Issue #1579, opened 2020-06-05
Desktop testing bugs


willmcvay:
**Describe the bug**

_Bugs came out of testing on desktop app. Should be tested in desktopMode as we as web_

- The client side welcome wizard shows in desktop, this should only be rendered in web mode.
- Uninstalling an app does not show a notification as before. The action of confirming the uninstall should trigger a refresh of the page in desktop mode as before using an ACButton component
- The same is true installing an app, it does not show a dialogue or trigger a refresh in desktop
- Uninstalling an app does not trigger a page refresh so you can repeatedly try to uninstall an app that is already uninstalled.
- The back to apps button on the apps pages should re-direct back to where you came from. So if I arrived on the app page from manage apps I should return to manage apps. Currently it will return you always to the browse apps page.

nphivu414:
Hi @willmcvay ,
I cannot reproduce the issue that doesn't trigger a page refresh after installing/uninstalling an app in desktop.
Please check my screen record:
![Untitled_ Jun 5, 2020 4_30 PM](https://user-images.githubusercontent.com/22409039/83861860-59d8fd00-a74b-11ea-8102-8ff8d0112dde.gif)

Thanks

PR #1612, opened 2020-06-08
fix: #1579 Desktop bugs on standalone app page

nphivu414:
# Pull request checklist

fixes: #

- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #1579 

## What is the new behavior?

- Fix issues that cause the desktop app not refreshing app list after uninstall/install an app

- [ ] Yes
- [x] No

### DIFF
--- a/packages/marketplace/src/components/ui/client-app-detail/client-app-install-confirmation.tsx
+++ b/packages/marketplace/src/components/ui/client-app-detail/client-app-install-confirmation.tsx
@@ -33,6 +33,7 @@ export const handleInstallAppSuccessCallback = (
         clientId,
       }),
     )
+    window.location.href = 'agencycloud://apps/refresh'
     closeInstallConfirmationModal()
     setIsSuccessAlertVisible(true)
   }
--- a/packages/marketplace/src/components/ui/client-app-detail/client-app-install-confirmation.tsx
+++ b/packages/marketplace/src/components/ui/client-app-detail/client-app-install-confirmation.tsx
@@ -12,6 +12,8 @@ import CallToAction from '../call-to-action'
 import { selectClientId } from '@/selector/client'
 import routes from '@/constants/routes'
 import { selectInstallationFormState } from '@/selector/installations'
+import { selectIsDesktopMode } from '@/selector/auth'
+import { DESKTOP_REFRESH_URL } from '@/constants/desktop-urls'
 
 export type ClientAppInstallConfirmationProps = {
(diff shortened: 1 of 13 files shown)


packages/marketplace/src/components/ui/client-app-detail/client-app-install-confirmation.tsx
```diff
@@ -33,6 +37,9 @@ export const handleInstallAppSuccessCallback = (
         clientId,
       }),
     )
+    if (isDesktopMode) {
```
phmngocnghia, line 40: This is repetitive, can we create an util function. E.g. handleIsDesktopMode

packages/marketplace/src/components/ui/client-app-detail/client-app-uninstall-confirmation.tsx
```diff
@@ -33,8 +37,11 @@ export const handleUninstallAppSuccessCallback = (
         clientId,
       }),
     )
-    closeUninstallConfirmationModal()
+    if (isDesktopMode) {
```
phmngocnghia, line 40: See comment ^

packages/marketplace/src/utils/auth-route.ts
```diff
@@ -53,29 +53,43 @@ export function getDefaultPathByLoginType({
...
   isDeveloperFirstTimeLoginComplete?: boolean
   isClientFirstTimeLoginComplete?: boolean
+  isDesktopMode?: boolean
 }) {
   switch (loginType) {
     case 'ADMIN':
       return Routes.ADMIN_APPROVALS
     case 'DEVELOPER':
       if (clientLoginRouteMatch) {
+        if (isDesktopMode) {
```
tanphamhaiduong, line 70: @willmcvay The logic here very complicated. So could we do reject if user login wrong portal instead of redirect ?

nphivu414: Currently, we let a developer account access to both routes "/developer/login" and  "/client/login" (and vice versa) although he's already logged in and then we will call a function named `getDefaultPathByLoginType` in `login.tsx` to get a URL to redirect them to an appropriate place.

This makes our logic in `getDefaultPathByLoginType`  becomes more complicated especially when we need to handle `isDesktopMode` case

PR #2348, opened 2020-08-12
feat: #2144 Expand GraphQL Service service for Vendors entity

ant066:
# Pull request checklist

fixes: #2144

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #2144

- [ ] Yes
- [x] No


packages/graphql-server/src/resolvers/vendors/__mocks__/vendor.ts
```diff
@@ -0,0 +1,73 @@
+export const vendorMock = {
```
willmcvay, line 1: Can you add a type annotation to your stubs please - also, these are stubs not mocks, should live in a `__stubs__` folder - thx

packages/graphql-server/src/resolvers/vendors/__mocks__/vendors.ts
```diff
@@ -0,0 +1,96 @@
+export const vendorsMock = {
```
willmcvay, line 1: As above, can we type notate our mocks please

packages/graphql-server/src/resolvers/vendors/api.ts
```diff
@@ -0,0 +1,172 @@
+import qs from 'query-string'
```
willmcvay, line 1: Looks like this file is missing tests - unless I have missed them

packages/graphql-server/src/resolvers/vendors/__mocks__/vendor.ts
nphivu414: Hi @willmcvay ,
We'are all putting these files into `__mocks__` folders for all services in graphql-server. Should we move all of them to `__stubs__` now?

willmcvay: Ahh, sorry, I must have missed this. Yeah @nphivu414 we should do because the `__mocks__` folder in jest has a special purpose ie if I call `jest.mock('../foo')` it will look for the `foo.ts` file in the `__mocks__` folder. Could get confusing. Stubs really should live in their own folder for clearer separation of concerns.

It's not a big deal though. @tanphamhaiduong do you have a view on this?

nphivu414: Hi @willmcvay 
I've just spoken with him about this, it makes sense to move stubs to `__stubs__` folder. @ant066  will fix it for Vendors entity here and I will fix the rest

willmcvay: Thanks @nphivu414

packages/graphql-server/src/resolvers/vendors/api.ts
tanphamhaiduong: @willmcvay look like entire graphql-server and marketplace we don't write test for the call api function

willmcvay: Yes, we should - this will be a public service, needs to be extremely well tested. Will write it up @tanphamhaiduong

Issue #2146, opened 2020-07-21
The Elements V2 package needs a contribution guide / specification


willmcvay:
**Background context or User story:**

_For the elements v2 project, there should be a set of contribution and usage docs that supply guiding principles and hints for developers_

**Specification or Acceptance Criteria:**

- Should be accessible from the route of the project
- Should populate the index.mdx
- The DocsWrapper container needs refinement and styling to support stateful components

PR #2578, opened 2020-09-04
feat: #2146 Elements v2 spec

willmcvay:
# Pull request checklist

fixes: #2146 

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Build (`yarn build`) was run locally and any changes were pushed
- [ ] Lint (`yarn lint`) has passed locally and any fixes were made for failures
- [ ] Test (`yarn test`) has passed locally and any fixes were made for failures

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Documentation content changes

Issue Number: #2146 

## What is the new behavior?

- Elements v2 spec

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/elements-next/src/index.mdx
+++ b/packages/elements-next/src/index.mdx
@@ -7,10 +7,105 @@ route: /
 
 Welcome to the homepage for V2.* of the Reapit Elements UI library. 
 
-## Usage
+## Specification
 
-TBC
+### Background
 
-## Contributing
+Elements v1.x has been a useful excercise to get us off the ground in terms of development. V2 / Next is a full re-write that will be the basis of Elements UI lib for the foreseeable future. The intention is to address the following structural problems with the v1 lib.
 
-TBC
+- Elements is very bloated both in terms of CSS and JS. We have too much unused CSS from Bulma and way to many NPM packages to do simple things.
+- Bulma UI lib makes it hard to change things, we have to override too many styles.
+- Inconsistent naming conventions and API for components / component props
+- Lack of variablisation means we can't change styles globally
+- API not fully documented by Storybook meaning options / variants / props for components and styles are not clear to external developers.
+- Poor support for non-react users. Most of our developers do not use React and we need to ensure the stylesheet solution is as usable for developers as the React components are for React Users.
+- Utilities mixed in with React components - both a poor separation of concens and insufficient documentation for the utilities.
+
+### Specification
+
+- Elements is a UI only library. Utilities and other modules should be migrated to the `@reapit/utils` package and will be documented and released as a separate NPM package.
+- CSS should be handwritten, we will not import any 3rd party CSS so we have full control over styles unless absolutely necessary eg DatePicker.
+- We should keep NPM packages to an absolute bare minimum, no packages to achieve trivial tasks to reduce library bloat.
+- All components to be strongly typed and heavily tested
+- Code that can be migrated or repurposed from v1 should be however, every component should be viewed critically, improved, upgraded, tidied and made consistent.
+- Components should be flexible where possible, render children to allow styling for custom components is a desirable pattern rather than binding specific single-use behaviours to the UI
+- Props should be consistent, boolean and imperative modifiers to allow for clean markup. These booleans should map to classes, re-usable where possible.
+
+```
+isActive
+isFlex
+isFlexColumn
+isTextCentered
+isPrimary
+isDanger
+isFullWidth
+hasChildren
+```
+- The generic styles folder contains a lib of generic classes, these should be used and composed over writing new CSS
+- All classes should be prefixed with `el` to ensure we do not clash with third party libs eg `elIsFlex` in Linaria will generate the class `el-is-flex`.
+- Colors, margins, borders, fonts, padding, font-sizes and anything that can be vairablized should be. We want to be able to change a global variable and see the changes everywhere.
+- CSS should look the same as the Bulma equivient where the component is equivient. Eg an input or a button in v1 should look identical in v2. We are actively hiring a designer and the styles will change in time but we want this to be simple to do when the time comes.
+- The docs should be styled the same as Elements - Elements docs should look like Elements!
+- There should be far fewer components. We have several different ways of doing notifications - there should be a single approach. See below for list of components to migrate.
+- HTML output should be next to the mounted React Component and the Rendered Component in the docs in all cases.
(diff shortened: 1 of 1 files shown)


tanphamhaiduong (approved):

packages/elements-next/src/index.mdx
```diff
@@ -7,10 +7,105 @@ route: /
...
+- A much better looking and mobile friendly date picker
+
+#### Possibly
+
+WSYWIG Editor - currently we use:
+
+- Himalaya
+- Pell
+
+Google Maps component - currently we use but seem to offer little benefit:
```
tanphamhaiduong, line 104: Should we need to add google maps component to elements? I think no need to add it to elements because only geo-diary use it

packages/elements-next/src/index.mdx
```diff
@@ -7,10 +7,105 @@ route: /
...
+- Grid, Grid Items - should be based on native CSS Grid, basis 12 cols
+- Loader - inline and page centered variants
+- Maps - consider ditching React Wrappers for Google maps - it's trivial to load a map in a UseEffect hook and saves tonnes of JS
+- Menu - shoud be refactored to better support actions, icons and sub menus as a dropdown
+- Notification, Toast, ToastMessage, all combined - single module to notify users, simple implementation with variants. Current notification is the most flexible.
+- Modal, single, simple implementation
+- Pagination
+- Spreadsheet - tidy and refactor as required
+- Table - just a simple styled HTML table, we don't use any advanced functionality and if we need it, just use Spreadsheet
+- Form - to export Formik, to include form Headings and Helpers
```
nphivu414, line 79: I think we don't need to include Formik to the elements V2, just need to export form components like Input, Select, CheckBox,... It's better if users can choose the form framework they want, Formik is not the best for every case.

packages/elements-next/src/index.mdx
```diff
@@ -7,10 +7,105 @@ route: /
...
+- Container, for pages and top level elements - as per Section, super flexible with modifiers
+- Grid, Grid Items - should be based on native CSS Grid, basis 12 cols
+- Loader - inline and page centered variants
+- Maps - consider ditching React Wrappers for Google maps - it's trivial to load a map in a UseEffect hook and saves tonnes of JS
+- Menu - shoud be refactored to better support actions, icons and sub menus as a dropdown
+- Notification, Toast, ToastMessage, all combined - single module to notify users, simple implementation with variants. Current notification is the most flexible.
+- Modal, single, simple implementation
+- Pagination
+- Spreadsheet - tidy and refactor as required
+- Table - just a simple styled HTML table, we don't use any advanced functionality and if we need it, just use Spreadsheet
```
nphivu414, line 78: It seems to me that we should keep using react-table, there are still some cases we need to render a pretty complex table with collapsible sub-rows like this whereas using a spreadsheet is an overkill in this case.
image

Also, react-table is not a super heavy lib, it's well-maintainable, lightweight (5kb - 14kb+ depending on features we used), and can handle almost every case we need.

packages/elements-next/src/index.mdx
```diff
@@ -7,10 +7,105 @@ route: /
...
+#### Definitely
+
+- DayJS
+- Formik
+- Linaria
+- React Icons
+- React Datasheet
+- React Image Crop
+- React Transition Group
+- A much better looking and mobile friendly date picker
```
nphivu414, line 95: The date picker we're using (react-datepicker) is fine to me, we can customize the style for better looking instead of using its default styles.
It also comes with a modal version that we can show to mobile users
image

packages/elements-next/src/index.mdx
```diff
@@ -7,10 +7,105 @@ route: /
...
+- CSS should look the same as the Bulma equivient where the component is equivient. Eg an input or a button in v1 should look identical in v2. We are actively hiring a designer and the styles will change in time but we want this to be simple to do when the time comes.
+- The docs should be styled the same as Elements - Elements docs should look like Elements!
+- There should be far fewer components. We have several different ways of doing notifications - there should be a single approach. See below for list of components to migrate.
+- HTML output should be next to the mounted React Component and the Rendered Component in the docs in all cases.
+- All variants of components should be visible at the top of the page in the generated props section. They should all be editable at the bottom, although only the basic component without variants should be rendered by default.
+- Where we export multiple components from the same file, we should export them as modules eg `<Headings.Subheading>`
+- Should work in Edge (latest version before Chromium), Edge Chromium, Firefox, Chrome, Safari
```
vuhuucuong, line 53: Just a note, I think we should add `postcss` in the build process to include `https://github.com/postcss/autoprefixer`, so that all the styles will not break in different browser.
There's a few postcss pluggin you may want to check out, like [cssnano](https://cssnano.co/)
Look like linaria supported `PostCSS` it too 

Screen Shot 2020-09-07 at 11 07 00

trankhacvy (approved):

packages/elements-next/src/index.mdx
willmcvay: @tanphamhaiduong So I think we will need maps in the future, but agree, not necessary for the first version of Elements v2 - we can cross that bridge when we come to it, will remove

packages/elements-next/src/index.mdx
willmcvay: @nphivu414 So I left Formik in because I thought that was the team's preference. I agree with you, I think it would be nicer if we were Form Lib agnostic. Happy to take this out, would be interested if anyone else has a view though?

packages/elements-next/src/index.mdx
willmcvay: @nphivu414 So probably the bit I didn't explain was that we had some user feedback from a couple of developers that said the Elements Table was hard to use and a bit janky. Also, react-table has no typings which is not ideal. Maybe we just need to document it better? Or another lib? What are the other advanced features we use? If it's just nested tables, this would be simple to write ourselves right?

packages/elements-next/src/index.mdx
willmcvay: @nphivu414  I do think it's really ugly so we need to spend some time styling it to be more "elements" if we keep it. I was not aware of the modal view, it is better definitely. Happy to keep the lib if we can improve the UX maybe worth looking at options though

packages/elements-next/src/index.mdx
willmcvay: Yep, agreed, we should definitely do this.

Issue #3067, opened 2020-11-27
Payments app should have a micro service to issue API keys for the payments mailer


willmcvay:
**Background context or User story:**

_To make payments for non-Reapit Connect users, we need to issue API keys on an email payment request so that we can match the payment id against a valid request. To do this, there should be a service that issues a temporary key that can be stored against email, payment id and other information. If a valid key is passed to the server, the server should use a client credentials flow to retrieve the payment request from the platform_

**Specification or Acceptance Criteria:**

- Should have 3 endpoints, a POST, PATCH and a GET
- POST should generate a GUID and store apiKey, createdAt, expiresAt, customerId, paymentId, email
- GET should accept the API key and retrieve the meta data, checking the key is valid and return a payment by id from the payments service
- PATCH should accept the payments patch payload, plus the API key and allow an update of payments status on success or fail of a payment
- The service itself should have an API key issued by the API gateway to add a layer of authentication, so the endpoint is not fully public

PR #3234, opened 2021-01-05
feat: #3067 payments proxy service

willmcvay:
# Pull request checklist

fixes: #3067


packages/payments-service/serverless.yml
```diff
@@ -0,0 +1,107 @@
...
+        TableName: ${self:custom.env.DYNAMO_DB_TABLE_NAME}
+        AttributeDefinitions:
+          - AttributeName: apiKey
+            AttributeType: S
+        KeySchema:
+          - AttributeName: apiKey
+            KeyType: HASH
+        ProvisionedThroughput:
+          ReadCapacityUnits: 1
+          WriteCapacityUnits: 1
```
jblok, line 82: I've also specified 1 read unit and 1 write unit in the event-status-service. As far as I can tell, 1 read unit allows you to read up to 4kb per second and 1 write can write up to 1kb per second.

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html

Not sure what your throughput will look like for this? Can always tweak provisioned settings as necessary, or we can run this "on demand" where it's pay per read/write.

packages/payments-service/src/controllers/create-api-key.ts
```diff
@@ -0,0 +1,42 @@
...
+    })
+    logger.info('Creating API key', { traceId, apiKeyObject })
+
+    const result = await db.put(apiKeyObject)
+
+    logger.info('Successfully created API key', { traceId, result })
+
+    res.status(201)
+    res.json({
+      apiKey: apiKey,
```
jblok, line 31: can just do `res.json({ apiKey })`

packages/payments-service/src/core/router.ts
```diff
@@ -0,0 +1,13 @@
+import { AppRequest, AppResponse } from '@reapit/node-utils'
+import { Router } from 'express'
+import { createApiKey } from '../controllers/create-api-key'
+import { getPayment } from '../controllers/get-payment'
+import { updatePayment } from '../controllers/update-payment'
+
+const router = Router()
+
+router.post('/api-key', (req: AppRequest, res: AppResponse) => createApiKey(req, res))
```
jblok, line 9: You can do also do it like this if you wanted - I find a bit simpler to read but you might disagree!

`router.post('/api-key', createApiKey)`

jblok (approved):

packages/payments-service/src/core/server.ts
```diff
@@ -1,9 +1,10 @@
+import 'isomorphic-fetch'
 import express from 'express'
 import bodyParser from 'body-parser'
 import session from 'express-session'
```
jblok, line 4: Oh also, I've removed session middleware from event-status-service - I also copied it from the payments-server package! But I can't see how it'll be needed as we're building stateless services

Issue #3546, opened 2021-02-22
Create an automations-service to handle interaction with Twilio for spike


jblok:
**Background context or User story:**

In order to fullfil tickers 43, 44, and 45 from Un-CRM we will need a backend service to manage interaction with the Twilio API since the Twilio API requires HTTP Basic authentication that can't be securely used in an SPA. Also the logic for managing the Twilio conversation may grow as the spike develops

**Specification or Acceptance Criteria:**

- Scaffold the automations service using serverless, and the event-status-service as a template
- Install the Twilio SDK and add authentication keys for my account
- Start playing with the Twilio APIs

PR #3598, opened 2021-02-26
feat: #3546 - add the barebones of the automations-service

jblok:
# Pull request checklist

fixes: #


package.json
```diff
@@ -15,6 +15,7 @@
       "packages/developer-portal",
       "packages/elements",
       "packages/event-status-service",
+      "packages/automations-service",
```
willmcvay, line 18: Think this can me remove now right?

packages/event-status-service/src/controllers/automation/create.ts
```diff
@@ -0,0 +1,53 @@
+import { Response } from 'express'
+import uuidv4 from 'uuid/v4'
+import { logger } from '../../core/logger'
+import { AppRequest, stringifyError } from '@reapit/node-utils'
+import { db } from '../../core/db'
+import { Automation, generateAutomationItem } from '../../schemas/automation.schema'
+
+type Payload = {
+  clientCode: Automation['clientCode']
```
willmcvay, line 9: Feels like we are losing some strong typing here... I am guessing TS doesn't like Automation.clientCode because it's not a class instance. Would it make sense to do something like;

```ts
interface Foo {
    // blah blah
}

class Automation implements Foo {
   // blah
}
```
Then Payload can be replaced by Foo and we are typesafe. Can explain if this doesn't make sense or if I am missing something :-)

packages/event-status-service/src/controllers/automation/create.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export default async (req: AppRequest, res: Response) => {
+  const payload = req.body as Payload
+  const { traceId } = req
+
+  try {
+    logger.info('Create new automation...', { traceId, payload })
+
+    if ((req as any).user.clientCode !== payload.clientCode) {
```
willmcvay, line 22: Feels like `req.user.clientCode` would be better as `req.user?.clientCode` same with `payload?.clientCode` just to avoid potential run time errors

packages/event-status-service/src/controllers/automation/delete.ts
```diff
@@ -0,0 +1,45 @@
...
+  const id = req.params.id as string | undefined
+  const { traceId } = req
+
+  try {
+    logger.info('Deleting automation by id...', { traceId, id })
+
+    const item = generateAutomationItem({ id })
+    const result = await db.get(item)
+
+    if (result.clientCode !== (req as any).user.clientCode) {
```
willmcvay, line 17: See above about run time errors

packages/event-status-service/src/controllers/automation/get.ts
```diff
@@ -0,0 +1,44 @@
...
+  const id = req.params.id as string | undefined
+  const { traceId } = req
+
+  try {
+    logger.info('Getting automation by id...', { traceId, id })
+
+    const itemToGet = generateAutomationItem({ id })
+    const result = await db.get(itemToGet)
+
+    if (result.clientCode !== (req as any).user.clientCode) {
```
willmcvay, line 17: Seeing this again, is it possible to extend AppRequest a generic eg AppRequest<{user: Foo>> so you don't have to cast it as any?

packages/event-status-service/src/controllers/automation/update.ts
```diff
@@ -0,0 +1,62 @@
...
+  const payload = req.body as Payload
+  const { traceId } = req
+
+  try {
+    logger.info('Updating automation by id...', { traceId, id })
+
+    const itemToGet = generateAutomationItem({ id })
+    const retrievedItem = await db.get(itemToGet)
+
+    if (retrievedItem.clientCode !== (req as any).user.clientCode) {
```
willmcvay, line 24: As above

willmcvay (commented):
Just a couple of suggestions to improve type safety. That aside, looks good 👍

willmcvay (approved):
LGTM 👍

Issue #3695, opened 2021-03-10
Event Webhook


AshDeeming:
**Background context or User story:**

Build webhook endpoint to ingest events and kick off automations and conversations

Also need an endpoint to return decorated events - allowing the user to search by event status & timestamp - also adding conversations and actions

Including the ability to make amendments to the webhook and events list endpoints, so that it’s using the real events API from platform team and the real event types

PR #3746, opened 2021-03-16
feat: #3695 - first step in using real events api

jblok:
# Pull request checklist

- Use the real events API to fetch recent events, before decorating them with a status, actions, and messages
- In the JSOn reutrn combine these events, with events retrieved the "old way" - i.e eventStatuses queried first, then decorated with event bodies, actions, and messages
- Detach a phone number from an existing Twilio conversation (if found to be in one) where the event is a test event. This is so that the same phone number can always be involved in a new conversation for a test


willmcvay (approved):
LGTM 👍

Issue #54, opened 2020-01-22
Inputs, forms and labels in Elements v3.0


HollyJoyPhillips:
**Background**

_Elements UI library is being refactored to ensure better DX, UX and a more lightweight, flexible library with a more consistent API_

**Task**

- To include text, password, number, email, date, time, tel, checkbox, radio, text area, file, select
- Include a separate styled label component
- Include a separate styled form component
- Create component in line with overall specification TBC
- Should implement inputs from Figma styleguide
- Should be flexible to allow use with any React Form or validation library by way of passing props to component
- Should show each component as a molecule story, as well as a form layout example with all inputs in it

phmngocnghia:
Hi @willmcvay, I thinks there's no need to create specific variant for input type. For me, a prop named "type" seems to be enough

PR #3863, opened 2021-03-30
feat:#54 elementsv3 input

jblok:
Ticket #54 

- Add an `Icon` component with icons from Figma, and that uses the standard intent prop (Icons are SVG and imported as markup by react still for now. Will look at a CSS-only solution later, and also still need to look at including the relevant loaders in the webpack config)
- Add a moelcule component for `Input` and `Label`, with examples of various input types in the docs
- Add a component `InputGroup` that styles `Input`, `Label` and `Icon` nicely together; make it work for HTML/CSS only usage and also a shorthand version for React users

### Minor amends
- `Button` text to uppercase
- Make the outline of the button (when focussed) to be a darker version of the background
- Update the colour palette with colours from Figma


packages/elements/src/components-v3/InputGroup/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  id = generateRandomId(),
+  children,
+  ...rest
+}) => {
+  if (!children)
+    return (
+      <ElInputGroup className={className}>
+        <Input id={id} {...rest} />
+        {icon && <Icon icon={icon} />}
+        {label && <Label for={id}>{label}</Label>}
```
willmcvay, line 35: Is it worth including a validation component here too? Or have I missed something and validation is part of Input?

packages/elements/src/components-v3/InputGroup/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  } catch (e) {
+    return ''
+  }
+}
+
+export const InputGroup: React.FC<IInputGroup> = ({
+  icon,
+  label,
+  className,
+  id = generateRandomId(),
```
willmcvay, line 26: Can we memoize generateRandomId so the id doesn't change at each render? I think it's much more efficient for React's DOM diffing.

willmcvay (approved):
Happy to approve, just a couple of small things to think about but don't want to block merging if out of scope

willmcvay (approved):
LGTM 👍🏻

PR #3882, opened 2021-04-01
feat: #54 elementsv3-input-2

jblok:
- Add Select component. Add docs for this usage on its own and with the Label component
- Add input[type=checkbox], plus docs & styling for it's use within `InputGroup` 

Plus
- PR feedback
- Add input[type=radio], plus docs & styling for it's use within `InputGroup` 
- Add baseline docs for some of the utility classes, just so they get imported into storybook OK


packages/elements/src/components-v3/Input/__styles__/index.tsx
```diff
@@ -2,19 +2,42 @@ import { styled } from 'linaria/react'
 
 export const inputBg = '#ffffff'
 export const inputFocusBg = 'var(--color-grey-light)'
-export const inputBorderBottom = 'inset 0px -1px 0px #000000'
+export const inputShadow = 'inset 0px -1px 0px #000000'
```
willmcvay, line 5: Feels like this one should be variablised maybe?

willmcvay (approved):
Just one thing but will leave up to you - am thinking you may want to remove the input shadow in the theme

willmcvay (approved):
LGTM 👍🏻

PR #3885, opened 2021-04-06
refactor: added http status code enum for human readable status code returns 

bashleigh:
Refactor to add http status code enum for human readable status codes on return from res.status etc. Checkout `packages/events-service/src/types/http.status.enum.ts`

Suggestion: 

Use exception classes and middlewares to reduce the need of duplicated conditions such as unauthorized responses. 

Example:

```ts
export class HttpException extends Error {
  constructor(
    readonly statusCode: HttpStatusCodeEnum = HttpStatusCodeEnum.INTERNAL_SERVER_ERROR,
    readonly message?: string,
  ) {
    super(message || statusCode.toString());
  }
}

export class BadRequestException extends HttpException {
  constructor(message?: string) { // Optional message for option of adding `validation errors`
    super(HttpStatusCodeEnum.BAD_REQUEST, message || "Bad Request");
  }
}

export class UnAuthorizedException extends HttpException {
  constructor(message?: string) {
    super(HttpStatusCodeEnum.UNAUTHORIZED, message || "Unauthorized");
  }
}

// Throwing

(req: Request, res: Response): Promise<Response | never> => {
  if (!req.user || req.user?.clientCode !== payload.clientCode) {
    throw new UnAuthorizedException();
  }
}

// Handling
if (error instanceof HttpException) {
  res.status(error.statusCode)
  res.json({error: error.message, code: error.statusCode});
  return res;
}

```
> Didn't have the NPM token to run tests locally

### DIFF
--- a/packages/events-service/src/controllers/automation/create.ts
+++ b/packages/events-service/src/controllers/automation/create.ts
@@ -5,6 +5,7 @@ import { logger } from '../../core/logger'
 import { AppRequest } from '../../types/request'
 import { db } from '../../core/db'
 import { AutomationRequiredFields, generateAutomationItem } from '../../schemas/automation.schema'
+import { HttpStatusCodeEnum } from '@/types/http.status.enum'
 
 type Payload = AutomationRequiredFields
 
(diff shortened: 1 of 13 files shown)


willmcvay (approved):
Thanks for this, LGTM 👍🏻

willmcvay:
Merging as it's an external fork so need to incorporate into the code base to run tests

Issue #3907, opened 2021-04-08
There should be a Render HTML Component for Storybook


willmcvay:
**Background context or User story:**

_Currently we use Storybook's HTML plugin to render out HTML versions of React Components however, to give a better experience for non-React users, would be great if we can create a component that renders out the static markup in our Story files_

**Specification or Acceptance Criteria:**

- Should be able to render a React Story in an MDX file as static markup
- Should be styled using prism or similar
- Should be laid out correctly using prettier if necessary
- Should handle overflows by wrapping text
- Should support a clickable "Copy to clipboard" button
- Approach is flexible - we could consider using React Dom's `renderDomToStaticMarkup` method or as per the HTML plugin, just grab the html from the renderd react component by using query selector
- If the component works well, apply to all stories and create own package for it so we can open source as an NPM package
- If the above is true, document correctly in the README.md

PR #3989, opened 2021-04-19
feat: #3907 storybookhtml

jblok:
#3907

Creates a new component `RenderHtmlMarkup` that can be used in stories like so

`<RenderHtmlMarkup component="V3/Button" story="Default Usage" />`


willmcvay (approved):
Works really nicely 👌🏻

Issue #3978, opened 2021-04-16
Website URL validation not allowing '-'


HollyJoyPhillips:
**Bug**: The website URL validation applied on the 'Website' field on the Organisations tab does not allow for '-'s. 

**Location**: https://developers.reapit.cloud/settings/organisation

**To replicate**: 
Add address 'https://www.holly-homes.com

![image](https://user-images.githubusercontent.com/52408607/115035133-27984000-9ec4-11eb-8d9a-2f23f2c78e8b.png)

**Without (-):**
![image](https://user-images.githubusercontent.com/52408607/115035198-3bdc3d00-9ec4-11eb-9c11-9c25e28f2bbc.png)

PR #3997, opened 2021-04-20
fix: #3978 fix website regex

willmcvay:
# Pull request checklist

fixes: #3978


packages/developer-portal/src/components/pages/settings/settings-organisation-tab/organisation-form/form-schema/validation-schema.ts
```diff
@@ -37,7 +37,13 @@ export const companyInformationFormSchema = Yup.object().shape({
   [websiteField.name]: Yup.string()
     .trim()
     .required(FIELD_REQUIRED)
-    .matches(websiteRegex, websiteField.errorMessage)
+    .test({
+      message: websiteField.errorMessage,
+      test: (value) => {
+        if (!value) return true
```
Shaun-Farrell, line 43: Just double checking this returns true if value is falsy.
Is this just because it gets caught by the required condition anyway?

willmcvay: In the instance that there is no value, it will be caught by Yup.string().required() and display the 'required' error. In this case, I don't want to override this message by displaying the website validation message at the next step, hence return true if an empty string. Vaguely make sense? :-)

Shaun-Farrell: Yes makes sense, all good then, I'll add approval.

Shaun-Farrell (approved):

Issue #2626, opened 2020-09-08
Loader in Elements v3.0


willmcvay:
**Background**

_Elements UI library is being refactored to ensure better DX, UX and a more lightweight, flexible library with a more consistent API_

**Task**

- Elements v3 should have a new loader component as per Figma designs
- Should support full page vertically and horizontally centred versions as well as an inline loader

PR #4021, opened 2021-04-22
feat: #2626 - loader tweaks

jblok:
- Adds an optional label to the loader
- Sets a sensible default width
- Increase the animation time


willmcvay (approved):
LGTM 👍🏻

Issue #3916, opened 2021-04-08
There should be a client config microservice for storing configs in parameter store


willmcvay:
**Background context or User story:**

_We use @reapit/config-manager to handle encrypted configs in AWS Parameter store. Task is to create a wrapper around this to allow us to store external config and access via REST for use in our web / node apps_

**Specification or Acceptance Criteria:**

- Service should have GET / POST / DELETE methods
- Should use query strings that map to flags in the config manager docs https://github.com/reapit/foundations/tree/master/packages/config-manager
- Should be protected by a Cognito Authorizer
- App should only have permissions to access systems manager with a restricted IAM role to namespace 'Client'
- Should use as entity query a client code and this client code should be validated against the decrypted id token

PR #4052, opened 2021-04-27
feat: #3916 new config micro service changes following PR

Shaun-Farrell:
Detail as per issue below (required):

#3916

* New core structure for config micro service
* Added changes following PR feedback and rebased


willmcvay (approved):
LGTM 👍🏻

PR #4072, opened 2021-04-28
feat: #3916 use the client code as part of the SSM Key

Shaun-Farrell:
# Pull request checklist

#3916

* I exported connectSessionVerifyDecodeIdToken from connect-session.
* I used this function to get client code for use in the SSM Key in new config-service
* Configs can be now be stored as "Client/SBOX/PaymentConfig" for example.
* Added a service in PaymentsClient to use the new config-service (to be cont...)
* Updated tests


willmcvay (approved):
Nice, LGTM 👍🏻

Issue #2633, opened 2020-09-08
Table in Elements v3.0


willmcvay:
**Background**

_Elements UI library is being refactored to ensure better DX, UX and a more lightweight, flexible library with a more consistent API_

**Task**

- Implement new table component in line with with Figma design for Payments app table
- Should have various cell types as per design, allowing for "has-icon" and "has-title" type classes
- Should have an optional expandable section below each row for calls to action, using toggle arrow at row end
- Should collapse in mobile to a tile like component with each section being either a left, right or full width aligned cell
- Should use CSS Grid or Flexbox, not a table as the underlying element for better mobile responsiveness

PR #4081, opened 2021-04-30
feat: #2633 elv3 table

jblok:
#2633

- Adds a Table component, and many smaller molecule components. Uses CSS grid to layout the contents.
- Supports a narrow view for smaller screen sizes, where it's a 2 column layout and cells can get special treatments applied.
- Adds styles, and lots of docs for the above.
- Fixes a typescript bug with the InputGroup. Base HTML type was incorrect.

Things not covered by this PR that might be useful or needed to use the table in prod

- The expandable row doesn't currently animate
- There aren any tests for the molecule components and not very thorough tests for the Table as a whole
- In the design one of the cells has a little status circle
- A prop or setting to allow a different order of cells on the narrow table view
- Modifiers for the `TableHeader` to allow it show show as ASC or DESC arrow
- Adding more breakpoints in, and allowing a configurable breakpoint (currently it's hard-set at <750px for narrow view)

And a possible refactor
- Perhaps `TableCell` shouldn't be supporting the special `darkText` prop and instead just make that the responsibility of the children it renders. E.g. it could just render something like `<p className="dark-text">...` which would reduce the complexity on `TableCell` (same thing could be done for the little status circles)


willmcvay (approved):
Approving as partial fix - will raise new ticket to address PR comments

Issue #4094, opened 2021-05-04
The Marketplace Management app should not use users PUT endpoint for Group Management


willmcvay:
**Background context or User story:**

_Currently we add and remove users from their groups using the `/users/:id` endpoint PUT. This endpoint will be deprecated so we should use POST `/groups/:id/members` and DELETE `/groups/:id/members/:memberId` to perform this task_ 

**Specification or Acceptance Criteria:**

- See https://github.com/reapit/foundations/issues/4064 for the back end ticket
- See: https://github.com/reapit/foundations/blob/master/packages/marketplace-management/src/components/ui/users/edit-user.tsx#L48 for the component that handles calls to the service

PR #4173, opened 2021-05-17
refactor: #4094 deprecated mm app put users endpoint

willmcvay:
# Pull request checklist

fixes: #4094


packages/marketplace-management/src/components/ui/users/edit-user.tsx
```diff
@@ -45,13 +43,35 @@ export const formFields: Record<FieldType, FormFieldInfo> = {
   },
 }
 
+export const sortAddRemoveGroups = (editingUser: UserModel, groupIds: string[]) => {
+  const currentGroups = editingUser.groups ?? []
+  const removeIds = currentGroups.filter((group) => !groupIds.find((groupId) => groupId === group))
```
bashleigh, line 48: use includes instead of reduce foreach
```suggestion
  const removeIds = currentGroups.filter((group) => !groupIds.includes(group))
```

packages/marketplace-management/src/components/ui/users/edit-user.tsx
```diff
@@ -45,13 +43,35 @@ export const formFields: Record<FieldType, FormFieldInfo> = {
...
+  const addIds = groupIds.filter((group) => !currentGroups.find((groupId) => groupId === group))
```
bashleigh, line 49: as as before
```suggestion
  const addIds = groupIds.filter((group) => !currentGroups.includes(group))
```

packages/marketplace-management/src/components/ui/users/edit-user.tsx
```diff
@@ -45,13 +43,35 @@ export const formFields: Record<FieldType, FormFieldInfo> = {
...
+  const { groupIds } = params
+  const userId = editingUser?.id
+  if (!groupIds || !editingUser || !userId) return null
+
+  const { removeIds, addIds } = sortAddRemoveGroups(editingUser, groupIds)
+  const totalUpdates = removeIds.length + addIds.length
+
+  const updateUserRes = await Promise.all([
+    ...removeIds.map((id) => removeMemberFromGroup({ id, userId })),
+    ...addIds.map((id) => addMemberToGroup({ id, userId })),
```
bashleigh, line 69: Batch endpoint?

bashleigh (changes requested):
Just some small comments on reducing complex looping and removing hard hitting API queries. Let me know what you think and if you want to discuss. Might want to check I've got the correct logic!

packages/marketplace-management/src/components/ui/users/edit-user.tsx
willmcvay: There should be... but there isn't...

In practice, there are only like 5 groups so it's not something to get too worried about!

bashleigh: ahhh ok fair 👍

bashleigh (approved):

Issue #4162, opened 2021-05-14
Deployment service


bashleigh:
Build a basic deployment serivce that manages deployments and enables their relationships with pipelines

Issue #4160, opened 2021-05-14
There should be a pipeline builder


willmcvay:
**Background context or User story:**

_For the Troy project, there should be a configurable, executable pipeline_

**Specification or Acceptance Criteria:**

- Should handle node or web apps
- Should retrieve a bundle / ball / config from S3 and execute build steps
- Should use Cloudform to spin up infra as appropriate to app type
- Should set up lambda, gateway and logging for node
- Should set up S3 & cloudfront for Web

PR #4219, opened 2021-05-25
feat: #4160 #4162 deployment service

bashleigh:
# Pull request checklist

fixes: #4160 
fixes: #4162


packages/api-key-service/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "api-key",
+  "dependencies": {
+    "@aws/dynamodb-data-mapper-annotations": "^0.7.3",
+    "@homeservenow/serverless-aws-handler": "^0.2.4",
+    "class-transformer": "^0.4.0",
+    "class-validator": "^0.13.1"
+  },
+  "scripts": {
```
willmcvay, line 9: Will expect a build, release and publish script too - you can echo skip them obvs, just makes the pipeline simple ie "run these tasks in order"

packages/deployment-service/src/utils/decode.token.ts
```diff
@@ -0,0 +1,10 @@
+import jwt from 'jsonwebtoken'
+
+export const decodeToken = (
```
willmcvay, line 3: We already have one of these - it also validates the token! See above

packages/api-key-service/src/functions/create.api-key.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ApiKeyDto } from '@/dto'
+import { ApiKeyModel } from '@/models'
+import { createApiKey as create } from '@/services'
+import { authorised, decodeToken } from '@reapit/node-utils'
+
+export const createApiKey = httpHandler<ApiKeyDto, ApiKeyModel>({
+  serialise: {
+    input: (event): ApiKeyDto => {
+      authorised(event)
+      const customer = decodeToken(event.headers['reapit-connect-token'] as string)
```
willmcvay, line 13: For decoding tokens you can use this https://github.com/reapit/foundations/blob/master/packages/connect-session/src/utils/verify-decode-id-token.ts#L79

Usage `import {connectSessionVerifyDecodeIdToken } from '@reapit/connect-session'`

willmcvay: Returns this interface https://github.com/reapit/foundations/blob/master/packages/connect-session/src/types.ts#L22

packages/cli/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "devDependencies": {
+    "@types/figlet": "^1.5.1",
+    "@types/inquirer": "^7.3.1",
+    "@types/yargs": "^16.0.2",
+    "concurrently": "^6.1.0",
+    "eslint": "^7.26.0",
+    "prettier": "^2.3.0",
+    "ts-node": "^9.1.1"
+  },
+  "scripts": {
```
willmcvay, line 23: As above, pipeline will expect all commands to be implemented even if skipped

packages/cli/README.md
```diff
@@ -0,0 +1 @@
+# cli
```
willmcvay, line 1: Can we call the package something a bit more helpful eg `deployment-cli`? 😁

packages/deployment-service/src/functions/create.deployment.ts
```diff
@@ -0,0 +1,43 @@
+import { httpHandler, ValidationException } from '@homeservenow/serverless-aws-handler'
+import { DeploymentDto } from '@/dto'
+import { DeploymentModel } from '@/models'
+import * as service from '@/services/deployment'
+import { plainToClass, classToClassFromExist } from 'class-transformer'
+import { validate } from 'class-validator'
+import { authorised } from '@/utils'
+import { decodeToken } from '@/utils/decode.token'
```
willmcvay, line 8: See above decoding tokens

packages/deployment-service/src/functions/create.pipeline.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { createPipeline } from './create.pipeline'
```
willmcvay, line 1: It's a small point but can tests live in a `__tests__` folder please? It's just the convention in the repo

packages/deployment-service/src/functions/paginate.deployment.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Return pagination response for signed in user
+ */
+export const paginateDeployments = httpHandler({
+  serialise: {
+    input: (event) => {
+      authorised(event)
+    },
+  },
+  handler: async ({ event }): Promise<QueryPaginator<DeploymentModel>> => {
+    const customer = decodeToken(event.headers['reapit-connect-token'] as string)
```
willmcvay, line 17: Decode token etc

tsconfig.json
```diff
@@ -4,8 +4,8 @@
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "jsx": "react",
-    "lib": ["dom", "es2015", "es2017", "es2018", "es2019", "es2020", "es5", "es6", "es7", "esnext"],
-    "module": "es2020",
+    "lib": ["dom", "es2015", "es2017", "es2018", "es2019", "es5", "es6", "es7", "esnext"],
```
willmcvay, line 7: Did you mean to remove this? Should tidy up the comment if so - be wary, may have effects on other packages!

packages/utils/src/node/authorisation/decode.token.ts
```diff
@@ -0,0 +1,10 @@
+import jwt from 'jsonwebtoken'
+
+export const decodeToken = (
```
willmcvay, line 3: Like I say, I don't think we need this func - we should validate tokens too and the existing one does this

willmcvay (changes requested):
So just a general comment - quite light on tests - suggest we add to the ticket list to review later -  don't want to block this getting merged

packages/deployment-service/src/utils/decode.token.ts
bashleigh: deleted

packages/api-key-service/src/functions/create.api-key.ts
bashleigh: deleted

packages/cli/README.md
bashleigh: was thinking reapit-cli

packages/deployment-service/src/functions/create.deployment.ts
bashleigh: deleted

packages/api-key-service/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "api-key",
```
willmcvay, line 2: Should be api-key-service - yarn uses the name for workspaces

packages/cli/README.md
```diff
@@ -0,0 +1,4 @@
+# reapit
```
willmcvay, line 1: You were going to change the package name from cli to developer-cli

packages/api-key-service/package.json
willmcvay, line 2: Think this needs to be api-key-service to match the directory - it's to do with how yarn workspaces work

willmcvay (changes requested):
Just a couple of small things...

packages/api-key-service/package.json
willmcvay: Don't know why but this comment is duplicated!

bashleigh: sure, I'll change it, did think this when I was watching the tests

willmcvay (approved):
💯

bashleigh:
100 commits exact too 😂

PR #4241, opened 2021-05-26
fix: added serverless setup for api key service

bashleigh:
# Pull request checklist

fixes: #


packages/api-key-service/serverless.yml
```diff
@@ -0,0 +1,132 @@
+service: "cloud-api-key-service"
```
willmcvay, line 1: Not sure if this needs quotes 🤔

willmcvay (approved):
Maybe worth looking if those quotes will cause a problem, should be fine 🤞🏻

Issue #4212, opened 2021-05-25
Add helper on GEO Diary


HollyJoyPhillips:
**Summary:** 
We should include a notification on the home screen of GEO diary, to explain to the user how to save the app to their home screen. 

**Task:** 
- Add notification on the home screen on first load 
- Should set cookie when message is dismissed
- Text "Did you know, you can save GEO Diary to your home screen? Click here to find out how" 

On an iphone, link to help (link TBC) 
On an android, link to help (link TBC)

Issue #4130, opened 2021-05-11
Link GEO Diary Contact with AML (Online Check List) app


HollyJoyPhillips:
**Summary**: A requested feature in GEO Diary is to be able to identify the state of a contact's id check and to view more information using the AML app. 

**Task**: 
- Provide the ability to access AML app for a specific contact
- Provide visibility of the id check status on the appointment tile (different colour icon?)
- Should default to normal GEO Diary visual if AML is not installed for that organisation 
- UI TBC

PR #4250, opened 2021-05-27
feat: #4130 #4212 adds aml checklist option to contact row

willmcvay:
# Pull request checklist

fixes: #4130 
fixes: #4212


packages/geo-diary/src/components/ui/contact-drawer/aml-check-row.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { FC } from 'react'
+import IconButton from '../icon-button'
+import Grid, { Col } from '../grid'
+import * as styles from './__styles__'
+import { Label } from '@reapit/elements/v3'
+
+interface AmlCheckContentProps {
+  contactId: string
+  name: string
```
bashleigh, line 9: Should this be string or undefined? Istead of empty string?

bashleigh (approved):

packages/geo-diary/src/components/ui/contact-drawer/aml-check-row.tsx
willmcvay: It doesn't really matter in this case - either way I will render it's container even without content so it doesn't break the flow. TBH, it's _very_ unlikely we would see a contact record without a name!

PR #4266, opened 2021-06-02
fix: added config to deployment webpack conf

bashleigh:


packages/api-key-service/webpack.config.js
```diff
@@ -5,6 +5,7 @@ const CopyPlugin = require('copy-webpack-plugin')
 webpackConfigNode.entry = slsw.lib.entries
 webpackConfigNode.optimization.minimize = slsw.lib.webpack.isLocal ? false : true
 
-webpackConfigNode.plugins.push(new CopyPlugin({ patterns: [{ from: 'publicKeys.json', to: PATHS.output }] }))
+webpackConfigNode.plugins.push(new CopyPlugin({ patterns: [{ from: 'public-keys.json', to: PATHS.output }] }))
```
willmcvay, line 8: Think you need to combine these two lines ie add both configs to the same patterns array - same below

willmcvay (changes requested):

packages/api-key-service/webpack.config.js
bashleigh: oh ok!

willmcvay (approved):
👍🏻

Issue #4309, opened 2021-06-09
There should be a test UI for the deployments service


willmcvay:
**Background context or User story:**

_Deploy new web app that allows us to test the deployment end to end flow_

**Specification or Acceptance Criteria:**

- Should allow a user to post to create an API key
- Should allow a user to retrieve their existing API keys
- Should list current deployments and any meta info eg pipelines 
- Anything else that makes sense!

PR #4319, opened 2021-06-10
feat: deployment service UI pages

bashleigh:
# Pull request checklist

fixes: #4309 , #4321

### DIFF
--- a/packages/api-key-service/src/functions/paginate.ts
+++ b/packages/api-key-service/src/functions/paginate.ts
@@ -13,6 +13,9 @@ type Pagintation<T> = {
 }
 
 export const paginateApiKeys = httpHandler<void, Pagintation<ApiKeyModel>>({
+  defaultOutputHeaders: {
+    'Access-Control-Allow-Origin': '*',
+  },
   handler: async ({ event }): Promise<Pagintation<ApiKeyModel>> => {
     let customer: LoginIdentity | undefined
 
--- a/packages/deployment-service-ui/package.json
+++ b/packages/deployment-service-ui/package.json
@@ -22,7 +22,6 @@
   },
   "dependencies": {
     "@reapit/connect-session": "workspace:packages/connect-session",
-    "@reapit/elements": "workspace:packages/elements",
     "@reapit/foundations-ts-definitions": "workspace:packages/foundations-ts-definitions",
     "@sentry/browser": "^5.30.0",
     "dayjs": "^1.10.4",
--- a/packages/deployment-service-ui/package.json
+++ b/packages/deployment-service-ui/package.json
@@ -32,6 +32,7 @@
     "react-ga": "^2.7.0",
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
+    "shleemy": "^1.2.2",
     "snyk": "^1.463.0",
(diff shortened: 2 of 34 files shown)


packages/api-key-service/src/functions/paginate.ts
```diff
@@ -13,6 +13,9 @@ type Pagintation<T> = {
 }
 
 export const paginateApiKeys = httpHandler<void, Pagintation<ApiKeyModel>>({
+  defaultOutputHeaders: {
```
willmcvay, line 16: Should the default headers live in a constant?

packages/deployment-service-ui/package.json
```diff
@@ -32,6 +32,7 @@
     "react-ga": "^2.7.0",
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
+    "shleemy": "^1.2.2",
```
willmcvay, line 35: Lol

packages/deployment-service-ui/src/components/pages/api-key/list.tsx
```diff
@@ -0,0 +1,110 @@
...
+      if (serviceResponse) {
+        setApiKeys(serviceResponse)
+      }
+    }
+    if (connectSession) {
+      fetchApiKeys()
+    }
+  }, [connectSession])
+
+  const createApiKey = async () => {
```
willmcvay, line 36: Usually I would say break these methods out so they can be tested but this is a quick and dirty app so no worries

packages/deployment-service-ui/src/components/pages/api-key/list.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+                return <span>{int.forHumans}</span>
+              },
+            },
+            {
+              id: 'Delete',
+              Cell: ({ row }: { row: { original: any } }) => (
+                <Button
+                  loading={deletionLoading.includes(row.original.id)}
+                  onClick={() => deleteApiKey(row.original.id)}
```
willmcvay, line 98: Again, usually I would avoid arrow functions onClick but in this case I don't care!

packages/deployment-service-ui/src/components/pages/deployments/list.tsx
```diff
@@ -0,0 +1,101 @@
...
+      if (serviceResponse) {
+        setDeployments([...deployments, ...serviceResponse])
+      }
+    }
+    if (connectSession) {
+      fetchDeployments()
+    }
+  }, [connectSession])
+
+  const deleteDeployment = async (id: string) => {
```
willmcvay, line 37: As per above, would break these out into handlers but don't worry about it for this app

packages/deployment-service-ui/package.json
bashleigh: ganna add it everywhere until it gets 2000 downloads a week :joy:

packages/api-key-service/src/functions/paginate.ts
bashleigh: yea why not!

willmcvay (approved):
LGTM 👍🏻

Issue #4324, opened 2021-06-11
The TS-Definitions workflow is broken 


willmcvay:
**Background context or User story:**

_See: https://github.com/reapit/foundations/actions/runs/926877225 - failing on test run_

**Specification or Acceptance Criteria:**

PR #4325, opened 2021-06-11
fix: #4324 fixes ts-definitions cron

willmcvay:
# Pull request checklist

fixes: #4324

### DIFF
--- a/packages/foundations-ts-definitions/scripts/create-index-file.js
+++ b/packages/foundations-ts-definitions/scripts/create-index-file.js
@@ -15,7 +15,7 @@ module.exports = () => {
         '// @ts-ignore\n' +
         'export * from "../deployment-schema"' +
         '// @ts-ignore\n' +
-        'export * from export * from "../api-key-schema"' +
+        'export * from "../api-key-schema"' +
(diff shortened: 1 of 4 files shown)


packages/foundations-ts-definitions/scripts/create-index-file.js
```diff
@@ -15,7 +15,7 @@ module.exports = () => {
         '// @ts-ignore\n' +
         'export * from "../deployment-schema"' +
         '// @ts-ignore\n' +
-        'export * from export * from "../api-key-schema"' +
+        'export * from "../api-key-schema"' +
```
bashleigh, line 18: Just so I know, how did this happen?

bashleigh: because of the @ts-ignore?

bashleigh (approved):

packages/foundations-ts-definitions/scripts/create-index-file.js
willmcvay: I screwed up

Issue #4034, opened 2021-04-23
Tile in Elements v3


willmcvay:
**Background context or User story:**

_As per general Elements v3 spec, add a flexible tike component to the v3 lib_

**Specification or Acceptance Criteria:**

- Should be based on the App tile in the App list view
- Should be flexible enough to use with other content
- Should have optional icon/image section
- Should be fixed height so tiles will align when on a grid
- Should have optional footer and header banners as per design

PR #4389, opened 2021-06-17
feat: #4034 elements tile

willmcvay:
# Pull request checklist

fixes: #4034


packages/elements/src/components-v3/Card/card-components.tsx
```diff
@@ -0,0 +1,220 @@
...
+        ))}
+      </CardContextMenuItems>
+    </CardContextMenuWrapper>
+  )
+}
+
+export const Card: FC<CardProps> = ({
+  className,
+  isMainCard,
+  isListCard,
```
bashleigh, line 129: should this be changed to `cardType: 'isMainCard' | 'isListCard'`? As the current condition relies on one of these being set. I'm finding it a little confusing to follow? In the tests I can see both being used? But in the component if neither is set then no children are visable. Is this correct?

bashleigh (commented):
Looks really good! Only one question about implementation that I'm not sure on or am not following

packages/elements/src/components-v3/Card/card-components.tsx
willmcvay: Yes, it can be either or both - worth looking at the designs :-)

bashleigh: thanks!

bashleigh (approved):

Issue #4410, opened 2021-06-21
Geo Diary final UI changes


willmcvay:
**Background context or User story:**

_Final UI changes ahead of re-launch - see Figma design for details_

**Specification or Acceptance Criteria:**

PR #4419, opened 2021-06-22
feat: #4410 final geo diary UI tweaks

willmcvay:
# Pull request checklist

fixes: #4410 

Still needs some tests and a tidy up but worth having an initial look @bashleigh


packages/elements/src/components-v3/Card/__styles__/index.ts
```diff
@@ -254,8 +262,14 @@ export const ElCardListItem = styled.div`
 `
 
 export const ElCardListItemTextWrap = styled.div`
+  font-size: 0.85rem;
   display: flex;
   flex-direction: column;
+  cursor: pointer;
+
+  ${isTablet} {
```
bashleigh, line 270: <3

packages/elements/src/components-v3/PersistantNotification/__styles__/index.tsx
```diff
@@ -26,13 +26,16 @@ export const elPnContent = css`
   padding: 0.75rem 1.25rem;
   opacity: 0;
   transition: 0.5s;
+  width: 100%;
 `
 
+export const elPnIsFullWidth = css``
+
+export const elPnIsFixed = css``
```
bashleigh, line 34: are these references?

packages/geo-diary/src/components/ui/map/google-map-component.tsx
```diff
@@ -381,9 +384,18 @@ export const GoogleMapComponent: FC<MapProps> = ({ appointments }) => {
...
+          mapTypeControl={false}
+          fullscreenControl={false}
+          zoomControlOptions={{
+            position: google?.maps?.ControlPosition.TOP_RIGHT,
+          }}
+          streetViewControlOptions={{
+            position: google?.maps?.ControlPosition.TOP_RIGHT,
+          }}
         />
+        {(isMobile || isTablet) && routeInformation && <TravelMode />}
```
bashleigh, line 396: Don't know if this is worth bringing up but do you think this condition should be reversed? `routeInformation && (isMobile || isTablet)`? Just thinking about what would likely be true or false to determine

bashleigh (approved):
LGTM!

Issue #4424, opened 2021-06-23
Create feature to notify cli user of updated version of cli available 


bashleigh:
add a notification in the cli to notify the user that there has been an update to the cli and they can download the updated version

PR #4431, opened 2021-06-23
feat: #4424 notify user of cli update

bashleigh:
# Pull request checklist

fixes: #4424 

I've added the feature and tested against @reapit/elements however as the cli is not currently published the axios request results in 406


willmcvay (approved):
LGTM 👍🏻

willmcvay (approved):

Issue #4089, opened 2021-05-04
Table component requires additional work


willmcvay:
**Background context or User story:**

_See https://github.com/reapit/foundations/pull/4081 there are a couple of outstanding issue on this PR to resolve_

**Specification or Acceptance Criteria:**

Specifically from the comments:

```
Things not covered by this PR that might be useful or needed to use the table in prod

The expandable row doesn't currently animate
There aren any tests for the molecule components and not very thorough tests for the Table as a whole
In the design one of the cells has a little status circle
A prop or setting to allow a different order of cells on the narrow table view
Modifiers for the TableHeader to allow it show show as ASC or DESC arrow
Adding more breakpoints in, and allowing a configurable breakpoint (currently it's hard-set at <750px for narrow view)
And a possible refactor

Perhaps TableCell shouldn't be supporting the special darkText prop and instead just make that the responsibility of the children it renders. E.g. it could just render something like <p className="dark-text">... which would reduce the complexity on TableCell (same thing could be done for the little status circles)
```

- The refactor suggestion looks to be the best way of dealing with the dark text and status circles
- Sorting and breakpoints we can leave as out of scope fow now
- Tests and animations are the most important tasks

bashleigh:
[ ] The expandable row doesn't currently animate
[ ] There aren any tests for the molecule components and not very thorough tests for the Table as a whole
[ ] In the design one of the cells has a little status circle
[ ] A prop or setting to allow a different order of cells on the narrow table view
[ ] Modifiers for the TableHeader to allow it show show as ASC or DESC arrow
[ ] Adding more breakpoints in, and allowing a configurable breakpoint (currently it's hard-set at <750px for narrow view)
And a possible refactor

[ ] Perhaps TableCell shouldn't be supporting the special darkText prop and instead just make that the responsibility of the children it renders. E.g. it could just render something like <p className="dark-text">... which would reduce the complexity on TableCell (same thing could be done for the little status circles)

PR #4470, opened 2021-06-30
refactor: table styling and documenting

bashleigh:
# Pull request checklist

fixes: #4089


packages/elements/src/components/table/__styles__/index.tsx
```diff
@@ -123,9 +271,41 @@ export const ElTableRow = styled.div`
...
+  &:hover,
+  &:focus {
+    box-shadow: 0px 0px 3px var(--color-blue-light);
+
+    .el-table-row {
+      box-shadow: none;
+    }
+  }
+
+  &${elIsActive} {
```
willmcvay, line 300: Should that not be `&.${elIsActive}` noting the `.`

packages/elements/src/components/table/__styles__/index.tsx
```diff
@@ -140,6 +320,8 @@ export const ElTableCell = styled.div`
     margin-right: 0;
   }
 
+  ${cellOrders}
+
   ${ElIcon} {
```
willmcvay, line 325: I don't think this works... I appreciate, not your add but worth checking

packages/elements/src/components/table/index.tsx
```diff
@@ -25,13 +27,14 @@ const getHeadersFromRows = (rows: Row[]): string[] => {
 export type Cell = {
   label: string
   value: string
-  children: React.ReactNode
+  children?: React.ReactNode
   icon?: IconNames
   statusCircleIntent?: Intent
   cellHasDarkText?: boolean
   narrowTable?: {
```
willmcvay, line 34: Is it worth defining `narrowTable` as it's own type?

packages/elements/src/components/table/table.stories.mdx
```diff
@@ -8,9 +8,16 @@ import {
...
+  TableRowContainer,
 } from './molecules'
 import { Icon } from '../icon'
 import { RenderHtmlMarkup } from '../../helpers/render-html-markup'
+import { Grid, Col, elSpan6, elOffset7 } from '../grid'
+import { Molecule } from '../layout'
+import { Subtitle } from '../typography'
+import { InputGroup } from '../input-group'
+import { Button } from '../button'
+import { StatusIndicator } from '../status-indicator'
```
willmcvay, line 20: A general comment but an `<ArgsTable component={Table} story="Foobar" />` component would be good on this page - one for each of the table cell and table components - just helps with understanding the available props

packages/elements/src/components/table/table.stories.mdx
```diff
@@ -148,6 +185,23 @@ If set, this prop will ensure that in narrow view mode the cell takes up both co
...
 
+### Narrow order 
+
+If you require to change the order of your cells in the mobile/narrow view, you can use the `narrowOrder` property to change the cell's order in the column.
+
+The order can be specified between 1 and 12.
+
+<Canvas>
+  <Story name="TableCell - Narrow order">
+    <TableCell narrowOrder={3}>
```
willmcvay, line 196: How do I do narrowOrder if I am non-React? Just a line of commentry - add this class etc

packages/elements/src/components/table/table.stories.mdx
```diff
@@ -218,6 +272,21 @@ Any other props given are applied to the underlying `div` element, e.g. `onClick
...
 
+#### TableExpandableRowTriggerCell narrowIsFullWidth 
+
+This property will force the trigger cell to be fullwidth of the table on narrow sized tables.
+
+> View in mobile view
+
+<Canvas>
+  <Story name="TableExpandableRowTriggerCell - narrowIsFullWidth">
+    <TableExpandableRowTriggerCell narrowIsFullWidth />
```
willmcvay, line 283: As per above, add this class if you're not using react

packages/elements/src/components/table/table.stories.mdx
```diff
@@ -395,6 +726,7 @@ type Row = {
   expandableContent: React.ReactNode
 }
 ```
+##### Inside cells array
 
 ```
 type Cell = {
```
willmcvay, line 732: If we have an ArgsTable, we can get rid of the inline types in the MD

willmcvay (changes requested):
Just a few small things - nice work though, that table's a pain!

packages/elements/src/components/table/__styles__/index.tsx
bashleigh: you mean ElIcon? I'll check it out

packages/elements/src/components/table/index.tsx
bashleigh: can do. Do you know if that argsTable thing works with Interfaces/Types?

packages/elements/src/components/table/table.stories.mdx
bashleigh: sure :)

packages/elements/src/components/table/table.stories.mdx
bashleigh: I couldn't get the argsTable component to work with interfaces or types. If it does work with those then happy days!

packages/elements/src/components/table/__styles__/index.tsx
bashleigh: I tried `&.${elIsActive}` but it didn't work. Plus what I wrote is used elsewhere in the file

packages/elements/src/components/table/__styles__/index.tsx
bashleigh: Screenshot 2021-06-30 at 16 22 20
this adds a margin right to the icon here

packages/elements/src/components/table/table.stories.mdx
bashleigh: I've added some but if you add an args table for components with no uniquely defined props `export const Example = () => ()` then you get some odd results like the previous component's props will appear in the args table

willmcvay (approved):

Issue #4448, opened 2021-06-29
Update label when requesting a feature from the Developer Portal (Help Page)


HollyJoyPhillips:
Task: Update the URL to set the pre-set label as 'external-request'

Issue #4436, opened 2021-06-24
Remove Live Chat section from 'Help'


HollyJoyPhillips:
**Task**: Remove the Live Chat section from within the 'Help' page in the developers portal

HollyJoyPhillips:
@willmcvay https://github.com/reapit/foundations/issues/4448

PR #4475, opened 2021-06-30
feat: #4436 #4448 remove live chat bot

willmcvay:
# Pull request checklist

fixes: #4436
fixes: #4448


packages/developer-portal/src/components/pages/help/help.tsx
```diff
@@ -40,57 +40,65 @@ export const handleFaq = (loginIdentity: LoginIdentity) => {
...
+    const liveChatItem = {
+      imgSrc: liveChatImg,
+      header: 'Need Help?',
+      text:
+        'If you have a question that is not covered in the documentation you can ask here. Please note ' +
+        'we don’t provide chat support for Agency Cloud Developer Edition.',
+      buttonText: 'START CHAT',
+      buttonOnClick: () => handleFaq(loginIdentity),
+    }
+    items.splice(4, 0, liveChatItem)
```
bashleigh, line 97: Are you able to do this using a different method? Using a key would be far neater and would avoid breaking the index required if someone say added a new array value that increased the index of the item you want to replace. 

So in stead of an array perhaps use an Object so that the key can be removed like so 

```ts
enum ItemKeys {
  LIVE_CHAT = 'live_chat',
}

const items: {[s: keyof  ItemKeys]: Item} = {
  [ItemKeys.LIVE_CHAT]: { ... },
  ...
}

items[ItemKeys.LIVE_CHAT] = { ... } // new item to replace with. 
//Or
delete items[ItemKeys.LIVE_CHAT]


// usage

Object.values(items).map(item => (<ItemComponent item={item} />))

```

Trying avoid a scenario where an item is added to the array altering the index of where you want to splice.

bashleigh (changes requested):
Looks good! Just a small suggestion/question regarding array indexes.

bashleigh (approved):

Issue #4407, opened 2021-06-20
Review all v3 components and refine documentation ahead of v3 release


willmcvay:
**Background context or User story:**

_Ahead of release of Elements v3, we need to ensure all components and docs are prod ready_

**Specification or Acceptance Criteria:**

- Some of our v3 components have changed in the styleguide since they were developed. 
- Some the docs can be improved and some need to be made consistent with the overall styleguide
- Address any code quality / consistency issues
- Ensure test coverage is as high as possible

PR #4482, opened 2021-07-01
feat: #4407 v3 component refactor

bashleigh:
# Pull request checklist

fixes: #4407 

Add floating button and button group components.

### DIFF
--- a/packages/elements/src/components/button/__styles__/index.tsx
+++ b/packages/elements/src/components/button/__styles__/index.tsx
@@ -12,7 +12,7 @@ import { elIsLoading } from '../../../styles/states'
 import { intentPrimary, intentSecondary, intentCritical, intentSuccess, intentDanger } from '../../../styles/variables'
 
 const buttonXPadding = 1.5
-const buttonYPaddingMobile = .25
+const buttonYPaddingMobile = 0.25
 const chevronLeft = (fill: string) =>
   `data:image/svg+xml;utf8,<svg width="18" height="40" viewBox="0 0 18 40" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M24 0L9.79882 0C8.09608 0 6.57999 1.07793 6.02073 2.6862L0.456861 18.6862C0.160976 19.5371 0.160976 20.4629 0.456861 21.3138L6.02073 37.3138C6.57999 38.9221 8.09608 40 9.79882 40H24V0Z" fill="${encodeURIComponent(
     fill,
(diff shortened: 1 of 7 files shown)


packages/elements/src/components/button/__styles__/index.tsx
```diff
@@ -113,13 +141,52 @@ export const ElButton = styled.button`
...
     }
   }
+
+  ${isMobile} {
+    height: 2.5rem;
+    font-size: 1rem;
+    padding: ${buttonYPaddingMobile}rem 0.75rem;
+  }
+
+  &.el-floating-button {
```
willmcvay, line 151: `.${elFloatingButton} {`

willmcvay: If the consuming app hashes the class name, `el-floating-button` won't be a thing!

packages/elements/src/components/button/__styles__/index.tsx
```diff
@@ -113,13 +141,52 @@ export const ElButton = styled.button`
...
+    justify-content: center;
+    align-items: center;
+
+    ${isMobile} {
+      width: 2.5rem;
+      height: 2.5rem;
+      margin: 0.5rem;
+    }
+
+    &${elIsLoading} {
```
willmcvay, line 168: Needs a `.` ie `&.${elIsLoading}`

packages/elements/src/components/button/__styles__/index.tsx
```diff
@@ -113,13 +141,52 @@ export const ElButton = styled.button`
...
+      &:after {
+        left: inherit;
+        top: inherit;
+      }
+
+      .el-icon {
```
willmcvay, line 174: As above, should be `.${elIcon}`

packages/elements/src/components/button/__styles__/index.tsx
```diff
@@ -217,3 +288,36 @@ export const elButtonHasRightChevron = css`
...
   }
 `
+
+export const ElButtonGroup = styled.div`
+  display: grid;
+  grid-auto-flow: column;
+  column-gap: 1rem;
+  row-gap: 1rem;
+
+  .el-button-size2 {
```
willmcvay, line 298: Same as above, the three button classes should be the raw class name

packages/elements/src/components/button/index.tsx
```diff
@@ -13,6 +14,18 @@ export interface IButton extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   chevronRight?: boolean
   fullWidth?: boolean
   className?: string
+  size?: 2 | 3 | 4
```
willmcvay, line 17: Should size be a type?

packages/elements/src/components/button/index.tsx
```diff
@@ -13,6 +14,18 @@ export interface IButton extends React.ButtonHTMLAttributes<HTMLButtonElement> {
...
+}
+
+const resolveButtonSize = (size: 2 | 3 | 4): string => {
```
willmcvay, line 20: Could use a test :-)

willmcvay (changes requested):

packages/elements/src/components/button/__styles__/index.tsx
bashleigh: I don't quite get this one. Sometimes I add a dot sometimes I don't 🤷‍♀️   in this case, if I add the dot, the applied styles disappear 🤷‍♀️

willmcvay (approved):

Issue #4035, opened 2021-04-23
Select Icon in Elements v3.0


HollyJoyPhillips:
**Background**

Elements UI library is being refactored to ensure better DX, UX and a more lightweight, flexible library with a more consistent API

**Task**
- Add interchangeable icon on the select component
- https://elements.reapit.cloud/?path=/story/v3-forms-select--default-usage

Issue #50, opened 2020-01-22
Support theming in Elements v3.0


HollyJoyPhillips:
**Background**

_Elements UI library is being refactored to ensure better DX, UX and a more lightweight, flexible library with a more consistent API_

**Task**

- Task in line with overall specification TBC
- Create comprehensive list of CSS variables to support theming that can be referenced across the styleguide and loaded globally
- Allow for an overrides theme to be added so users can customise the base variables to provide a theme

PR #4515, opened 2021-07-03
feat: #4407 elements prod refinements

willmcvay:
# Pull request checklist

fixes: #4407 
fixes: #50 
fixes: #4035


packages/elements/src/components/button/button.tsx
```diff
@@ -69,7 +69,7 @@ export interface IFloatingButton extends ButtonProps {
 export const FloatingButton: FC<IFloatingButton> = ({ icon, intent, ...rest }) => {
   return (
     <Button className={styles.elFloatingButton} intent={intent} {...rest}>
-      <Icon icon={icon} intent={'neutral' as Intent} iconSize="small" />
+      <Icon icon={icon} intent={intent ? 'neutral' : undefined} iconSize="small" />
```
bashleigh, line 72: <3

packages/elements/src/components/select/__styles__/index.ts
```diff
@@ -20,4 +20,8 @@ export const ElSelect = styled.select`
     outline: none;
     background-color: var(--component-input-focus-bg);
   }
+
+  &:hover {
+    background-image: url('data:image/svg+xml;utf8,<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M11.0498 9.82632L6.00274 4.44338L0.956931 9.82492C0.738006 10.0584 0.383056 10.0584 0.164132 9.82489C-0.0547151 9.59138 -0.0547118 9.21287 0.164149 8.97937L5.60635 3.17506C5.82528 2.94164 6.18021 2.94165 6.39913 3.17507L11.8413 8.97936L11.8426 8.98079C12.0577 9.21834 12.0516 9.59689 11.8288 9.82631C11.6116 10.0501 11.2671 10.0501 11.0498 9.82632Z" fill="currentColor"/></svg>');
```
bashleigh, line 25: Just a small comment, could be bas64 encode this data images? Would save a few bytes I reckon? Not entirely sure how many but would lower our file sizes for downloads

packages/elements/src/storybook/concepts.stories.mdx
```diff
@@ -0,0 +1,92 @@
...
+The various intent types are below and are consistent accross the library.
+
+| Intent    | Usage                                                                                                  |
+| --------- | ------------------------------------------------------------------------------------------------------ |
+| primary   | The main/most likely action a user might want to take                                                  |
+| secondary | Typically used alonside a primary button to offer an alternative action, e.g. a Cancel or Close button |
+| critical  | An action that should be the very next, or only logical action that a user can take                    |
+| success   | An action that would lead to a good or pleasant outcome                                                |
+| danger    | An action that would lead to a potentially bad or dangerous outcome, e.g a delete button               |
+| neutral   | An action that has no specific meaning - essentially no intent is inferred from                        |
```
bashleigh, line 88: Stupid suggestion. Could we add coloured emoji's for the intents? 
https://stackoverflow.com/questions/11509830/how-to-add-color-to-githubs-readme-md-file/65174367#65174367

packages/elements/src/storybook/intro.stories.mdx
```diff
@@ -6,52 +6,116 @@ Welcome to Reapit Elements UI Library.
...
 
-TODO - Expand these - just adding some notes for now
+### Installation
+
+Elements is distributed via NPM as a CommonJS, ESM and UDM module. Your first step is, with NodeJS installed, to run;
+
+```bash
+npm install @reapit/elements
+
+# or for, yan users
```
bashleigh, line 20:
```suggestion
# or for, yarn users
```

packages/elements/src/storybook/intro.stories.mdx
```diff
@@ -6,52 +6,116 @@ Welcome to Reapit Elements UI Library.
...
+import { Input } from '@reapit/elememts'
+
+// Or for CommonJS users
+
+const { Input } = require('@reapit/elements')
+```
+
+You will also need to add the exported stylesheet from the dist folder of the library, to the entry point of your application eg;
+
+```tsx
```
bashleigh, line 37: doesn't look to me that github does tsx? 
```suggestion
```ts
```

I could be wrong but not getting the highlights in the review

packages/elements/src/styles/box-alignment.ts
```diff
@@ -1,47 +0,0 @@
-import { css } from 'linaria'
```
bashleigh, line 1: Forgot to delete this sorry!

packages/elements/src/styles/globals.ts
```diff
@@ -177,18 +177,18 @@ export const elVariables = css`
...
-      --color-blue-light: ${colorBlueLight};
-      --color-blue-light2: ${colorBlueLight2};
-      --color-blue-dark: ${colorBlueDark};
-      --color-blue-dark2: ${colorBlueDark2};
-      --color-accent-orange: ${colorAccentOrange};
+      --color-blue-light: #23a4de;
+      --color-blue-light2: #7bc9eb;
+      --color-blue-dark: #262f69;
+      --color-blue-dark2: #31448f;
+      --color-accent-orange: #ffb71b;
```
bashleigh, line 184: why did you remove the vars? Just curious and if the consts are no longer used then perhaps they should be deleted too? Thinking about SRP => no duplicated vars

bashleigh (approved):
All good! Just a few comments, points, suggestions and questions :)

packages/elements/src/components/select/__styles__/index.ts
willmcvay: We could but I think that svg is only a few bytes!

willmcvay: Future enhancement - there are a few places we do this

packages/elements/src/styles/globals.ts
willmcvay: It's because I needed to be able to copy-paste the variables list into the docs when I update it and it wasn't obvious when I did this where the variables in the template literals came from - was easier to document this way, not ideal I appreciate

packages/elements/src/storybook/concepts.stories.mdx
willmcvay: Erm, maybe... TBH, not sure how they would look in Storybook though...

packages/elements/src/components/select/__styles__/index.ts
bashleigh: I'm just thinking about the hundreds of SVGs I've seen

packages/elements/src/storybook/concepts.stories.mdx
bashleigh: sure, was just an idea :)

packages/elements/src/styles/globals.ts
bashleigh: ahhh I see, I'm just worried that having the hex' declared in 2 places may become confusing if there were to be changes to colours. Like if you said to me "Ashleigh, please update the danger colour" I would likely find one variable and change that variable and likely not realise there are 2 separate declarations. Would we be able to delete the TS consts to avoid confusion or too much of a hassle?

PR #4598, opened 2021-07-15
feat: deployment executables

bashleigh:
General deployment executable function actions


packages/deployment-service/src/executables/pull.ts
```diff
@@ -1,9 +1,37 @@
...
+  zip.extractAllTo(`${dir}`, true)
+}
+
+export const pull: ExecutableType = async (task: TaskEntity, pipeline: PipelineEntity): Promise<true | never> => {
   console.log('pull...')
   console.log('executable', task)
 
+  try {
+    await new Promise<void>((resolve) => {
+      request(`https://codeload.github.com/${pipeline.repository}/zip/refs/heads/master`)
```
willmcvay, line 19: Be nice if this were a constant - the base uri that is

packages/deployment-service/src/executables/pull.ts
```diff
@@ -1,9 +1,37 @@
...
+        .pipe(fs.createWriteStream(`${dir}/${cloneZip}`))
+        .on('close', function () {
+          resolve()
+        })
+    })
+
+    // console.log(result)
```
willmcvay, line 26: Couple of comments

willmcvay (approved):
Just a couple of small bits

packages/deployment-service/src/executables/pull.ts
bashleigh: this'll all be updated and removed :)

Issue #4541, opened 2021-07-07
The CRA template needs updating to use Elements v3


willmcvay:
**Background context or User story:**

_See #4516 this requirement was missed - need to update the CRA to use Elements v3 and release to prod_

PR #4640, opened 2021-07-22
feat: #4541 cra template updates

willmcvay:
# Pull request checklist

fixes: #4541


packages/cra-template-foundations/template/src/index.tsx
```diff
@@ -30,8 +30,9 @@ export const renderApp = (Component: React.ComponentType) => {
 
 const run = async () => {
   try {
+    const globalConfig: Config = config
     // Set the global config
-    window.reapit.config = config as Config
+    window.reapit.config = globalConfig
```
bashleigh, line 35: do we need the delcare here?

bashleigh (approved):
LGTM! Only one thing I noticed. SVGs now have no EOF line? Guess new prettier conf will fix this?

packages/cra-template-foundations/template/src/index.tsx
willmcvay: Not sure why but TS didn't like it unless I added the type 🤷🏻

willmcvay:
> LGTM! Only one thing I noticed. SVGs now have no EOF line? Guess new prettier conf will fix this?

Those are auto generated, again, not sure why....

Issue #4654, opened 2021-07-26
Marketplace handling when using demo URL


HollyJoyPhillips:
Task: 
- Hide 'Apps' from navigation when URL is 'demo=true' - or user is hollyjoyphillips+res@gmail.com 
- When routing to AML from GEO Diary apply the same logic above

PR #4666, opened 2021-07-27
feat: #4654 hide apps button demo

willmcvay:
# Pull request checklist

fixes: #4654


joshbalfour (approved):

packages/aml-checklist/src/components/ui/menu.tsx
```diff
@@ -38,21 +41,21 @@ export const generateMenuConfig = (logoutCallback: () => void, location: Locatio
...
-      {
+      !isDemo && {
         title: 'Logout',
         key: 'LOGOUT',
         callback: logoutCallback,
         icon: <ProfileIcon />,
         type: 'SECONDARY',
       },
-    ],
+    ].filter(Boolean) as MenuItem[],
```
bashleigh, line 58: Can do a clever trick to do away with the cast. You'll probably decide against it but it's a way around the cast :) 

```ts
const isMenuItem = (menu: any): menu is MenuItem => typeof menu === 'object' && menu.hasOwnProperty('title')

return [menuItem, false].filter<MenuItem>(item => isMenuItem(item))
```

bashleigh (approved):
Looks good! Just a small suggestion to avoid casting

packages/aml-checklist/src/components/ui/menu.tsx
willmcvay: But you said the menu was of type any!

Issue #4605, opened 2021-07-16
create multiple pages


joshbalfour:
**Background context or User story:**
As a user I want to be able to create multiple pages, which I will later link to.

**Specification or Acceptance Criteria:**
Ability to create, edit, and delete pages

PR #4672, opened 2021-07-28
feat: app builder multi page

joshbalfour:
# Pull request checklist

fixes: #4605


packages/app-builder/src/components/ui/header/PageSelector.tsx
```diff
@@ -0,0 +1,39 @@
...
+      id="page_selector"
+      name="page_selector"
+      value={pageId}
+      onChange={(e) => {
+        onChange(e.target.value)
+      }}
+    >
+      {getPages()
+        .map((p) => ({ value: p.id, label: p.name }))
+        .map(({ value, label }) => (
```
bashleigh, line 19: double map? The below should rename the params 

```suggestion
        .map(({ id: value, name: label }) => (
```

packages/app-builder/src/components/ui/header/PageSelector.tsx
```diff
@@ -0,0 +1,39 @@
...
+          </option>
+        ))}
+      <option value="">Select a page</option>
+    </Select>
+    <Button
+      style={{ zoom: 0.78 }}
+      className={elM1}
+      onClick={() => {
+        const pageName = prompt('Page name?')
+        if (!pageName) return
```
bashleigh, line 31: eslint won't like this 
```suggestion
        if (!pageName) {
            return
        }
```

packages/app-builder/src/components/ui/header/emptyState.ts
```diff
@@ -0,0 +1,24 @@
+export default {
```
bashleigh, line 1: default export again? :p

bashleigh (approved):
Looks good! Just added some suggestions :)

Issue #4694, opened 2021-08-02
A basic subset of admin functionality should be permitted


willmcvay:
**Background context or User story:**

_In the developer admin portal, Reapit employees should be able to run basic admin reports from the portal_

PR #4719, opened 2021-08-04
feat: #4694 limit admin permissions for whitelist users

willmcvay:
# Pull request checklist

fixes: #4694


joshbalfour (approved):
Are we limiting endpoint accessibility on the server side also?

Issue #4523, opened 2021-07-06
There should be a custom query graphql service


willmcvay:
**Background context or User story:**

_To support our app builder project, there should be a graphql server that allows us to craft custom queries by aggregating and normalising platform data_

**Specification or Acceptance Criteria:**

- Task is to scaffold this as a Node application
- Should be authenticated using an idtoken and lambda authorizer, passing an access token downstream to the platfom
- Should use Lambda if possible and have a Serverless config
- Task should give some solid thought to how to structure code to make it scalable in the future
- Establish design patterns to handle business logic in resolvers and re-use API calls to platform
- Investigate the best GraphQL tooling - currently we use Apollo server but worth looking at whether this is still the best choice for us
- Should demonstrate one basic query in a GQL playground
- Should be deployed to AWS dev env

Issue #4604, opened 2021-07-16
Create generic dataset to be edited by frontend


joshbalfour:
**Background context or User story:**
e.g. Teams and Players, where a Player has multiple attributes which can be edited, and Teams can be created which are made up of players

**Specification or Acceptance Criteria:**
The backend holds both the dataset and the schema. The schema describes what kind of input the frontend should use to edit each property of the data, which fields are required, and which are optional.
The frontend components to use this API are out of scope for this ticket.

PR #4748, opened 2021-08-09
feat: app builder backend

joshbalfour:
A bit big as it's adding a whole new service.
Dockerfile is top-level as it's reusable across all packages and needs the context of the whole monorepo.

Thinking that we could later extract the cdk components used into a top level package which could be reused across the other packages, which cdk suits itself to quite nicely.

Fixes: #4523, #4604


packages/app-builder-backend/src/entities/Author.ts
```diff
@@ -0,0 +1,13 @@
+import { ObjectType, Field, ID } from 'type-graphql'
```
willmcvay, line 1: Tiny thing but kebab-case for files - it's not a big deal as it's only stub data so don't feel you have to change it

packages/app-builder-backend/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "main": "./src/index.ts",
+  "scripts": {
+    "start": "ts-node .",
+    "test": "echo '...skipping...'",
+    "build": "tsc",
+    "lint": "eslint --cache --ext=ts,tsx,js src --fix",
+    "release": "cdk deploy --require-approval never",
+    "publish": "yarn npm publish --access public",
+    "conf": "echo '...skipping...'",
+    "synth": "cdk synth",
```
willmcvay, line 20: What does this do? Just interested :-)

packages/app-builder-backend/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "type": "git",
+    "url": "git+git@github.com:reapit/foundations.git"
+  },
+  "license": "MIT",
+  "author": "Reapit Ltd",
+  "main": "./src/index.ts",
+  "scripts": {
+    "start": "ts-node .",
+    "test": "echo '...skipping...'",
+    "build": "tsc",
```
willmcvay, line 15: Does this include your node modules correctly? Great if we can get away with no webpack

willmcvay (approved):
LGTM 👍🏻

packages/app-builder-backend/package.json
joshbalfour: yep! the node modules get bundled into the docker container on build
https://github.com/reapit/foundations/blob/feat/%234523-app-builder-backend/Dockerfile#L50

packages/app-builder-backend/package.json
joshbalfour: it generates the cloudformation template from the cdk code and validates it, useful if you make a cdk code change and want to make sure that it'll deploy without having to actually deploy it

PR #4750, opened 2021-08-10
[WIP] feat: #4427 cdk deployment

bashleigh:
# Pull request checklist

fixes: #4427


packages/elements/src/components/status-indicator/__styles__/index.ts
```diff
@@ -44,4 +44,10 @@ export const ElStatusIndicator = styled.span`
     color: var(--intent-danger-text);
     outline-color: var(--intent-danger-dark);
   }
+
+  &.shape-tag {
```
willmcvay, line 48: Can we add `elShapeTag` as an exported class from elements and bump a minor version please?

willmcvay (commented):
Just one small thing - also, be good if @joshbalfour could have a look at the CDK code :-)

packages/elements/src/components/status-indicator/__styles__/index.ts
bashleigh: sure!

packages/deployment-service/src/utils/project-dir.ts
```diff
@@ -4,6 +4,9 @@ import { PipelineEntity } from './../entities'
 export const projectDir = (dir: string, pipeline: PipelineEntity) =>
   resolve(dir, `${(pipeline.repository as string).split('/').pop()}-master`)
 
-export const developerDir = (pipeline: PipelineEntity): string => `/mnt/efs1/project/${pipeline.developerId}/`
+const startDir = 'mnt/efs1'
+// const startDir = 'tmp'
```
joshbalfour, line 8: needed still?

joshbalfour (approved):
LGTM 😄

packages/deployment-service/src/utils/project-dir.ts
bashleigh: Sort of, use /tmp for local testing. Quite annoying but don't really want to add an env for this just yet

Issue #4618, opened 2021-07-16
fetch and render a collection of objects as a table


joshbalfour:
**Background context or User story:**
As a user I want to list a collection of objects and view them as a table

**Specification or Acceptance Criteria:**
As above

PR #4792, opened 2021-08-17
feat: app builder object list

joshbalfour:
# Pull request checklist

fixes:  #4618


packages/app-builder/src/components/hooks/objects/use-object.list.ts
```diff
@@ -0,0 +1,17 @@
+import { useQuery } from '@apollo/client'
+import { useIntrospection, dummyQuery } from '../use-introspection'
+
+const getOnlyKey = (obj: any) => obj[Object.keys(obj)[0]]
```
bashleigh, line 4:
```suggestion
const getOnlyKey = <T extends any, R extends any>(obj: T): R | undefined => obj[Object.keys(obj)[0]]
```

Is this to get the first entry of the object? `Object.values(obj)[0]` would be a better solution?

packages/app-builder/src/components/hooks/use-introspection/helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const getObjectType = (type: IntrospectionOutputTypeRef): string => {
+  if (isNonNullType(type) || isListType(type)) {
+    return getObjectType(type.ofType)
+  }
+
+  return type.name
+}
+
+export const notEmpty = <TValue>(value: TValue | null | undefined): value is TValue => {
```
bashleigh, line 20:
```suggestion
export const notEmpty = <TValue>(value?: TValue | null): value is TValue => {
```

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
```diff
@@ -0,0 +1,58 @@
+import { IntrospectionOutputTypeRef, IntrospectionObjectType } from 'graphql'
+import { flatKind, getObjectType } from './helpers'
+import { LIST, NON_NULL, OBJECT, QueryableField, SCALAR } from './types'
+
+type NestedDict = {
+  [key: string]: NestedDict | undefined
```
bashleigh, line 6:
```suggestion
  [key: string]?: NestedDict
```

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const nestedFieldsToString = (nestedFields?: NestedDict) => {
+  if (!nestedFields) {
+    return ''
```
bashleigh, line 11:
```suggestion
    return '{ }'
```

return empty object?

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (type.kind === SCALAR) {
+    return undefined
+  }
+
+  if (type.kind === OBJECT) {
+    const objectType = queryableObjectTypes.find(({ name }) => name === type.name)
+    if (objectType) {
+      objectType.fields.forEach((field) => {
+        nestedDict[field.name] = queryableFieldToNestedDict(field.type, queryableObjectTypes)
+      })
```
bashleigh, line 41:
```suggestion
      return objectType.fields.reduce<NestedDist, string>((dict, field) => ({
       ...dict,
       [field.name]: queryableFieldToNestedDict(field.type, queryableObjectTypes),
      }), nestedDict)
```

Might need to reverse the generic types + parameters - I always get reduce params/types the wrong way around

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  return nestedDict
+}
+
+export const getTopLevelFields = (
+  objectTypes: Array<IntrospectionObjectType>,
+  queryName: string,
+): Array<QueryableField> =>
```
bashleigh, line 51: My old boss used to kill me for not being consistent between `[]` and `Array<>`. I feel like I need to mention it but I really don't want to 😂 I'm not bothered but I can't not mention it now 😱

packages/app-builder/src/components/hooks/use-introspection/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  IntrospectionOutputTypeRef,
+  IntrospectionNonNullTypeRef,
+  IntrospectionListTypeRef,
+  IntrospectionField,
+} from 'graphql'
+
+export const OBJECT = 'OBJECT'
+export const NON_NULL = 'NON_NULL'
+export const LIST = 'LIST'
+export const SCALAR = 'SCALAR'
```
bashleigh, line 12: Could use an enum here? That way we can group them with a generic type of say `ObjectType` where `ObjectType` is one of these values

bashleigh: Comes in handy for types like `{[keyof ObjectType]: Function}`

packages/app-builder/src/components/ui/user/table.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const ObjectTableCell = ({ __typename }: any) => {
+  return <span>[{__typename} embedded object]</span>
+}
+
+const Table = ({ typeName, ...props }: TableProps) => {
+  const { data, loading } = useObjectList(typeName)
+  const rows =
+    data &&
+    data.map((row) => {
```
bashleigh, line 24: Could use reduce?

packages/app-builder/src/components/ui/viewport/index.tsx
```diff
@@ -56,10 +56,18 @@ const Viewport = ({ children, iframeRef }) => {
...
 
   useEffect(() => {
-    if (pageId) {
-      const page = getPage(pageId)
-      actions.deserialize(page.nodes)
+    let page
+    try {
+      page = getPage(pageId)
+    } catch (e) {
+      //
```
bashleigh, line 63: 😱

bashleigh (changes requested):
Added quite a few suggestions, guessing most can be ignored for now!

packages/app-builder/src/components/hooks/objects/use-object.list.ts
joshbalfour: brain fart, you're right 😅

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
joshbalfour: this generates any subfields which should go on to the graphql query, so either 
```
property
```
for a scalar field or
```
property {
  subproperty
}
```
for an object field, so if property has no sub properties you don't add anything, hence the empty string

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
joshbalfour: you got it right :D

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
joshbalfour: i consistently (i hope!) use `Array<any>` but happy to change

willmcvay (approved):
With Ashleigh's comments LGTM 👍🏻

packages/app-builder/src/components/ui/viewport/index.tsx
joshbalfour: 😭

packages/app-builder/src/components/hooks/use-introspection/nested-fields.ts
bashleigh: 👍

bashleigh (approved):
👍

Issue #4766, opened 2021-08-12
Updates to the  Admin Portal 


HollyJoyPhillips:
**Summary:** We should improve the filtering options/information rendered on the Admin portal and provide the ability to check against webhooks. 

**Task:** 
**Additional Column**
- Add ‘Private’ column on /apps 
- Should check against limitToClientIds, if populated display tick 

**Additional Filter**
- Add search field ‘Is Listed’ 
- Should check against isListed

**Check Webhooks**
- Add ‘Webhook’ button on apps
- When clicked should call https://platform.reapit.cloud/webhooks/subscriptions by app ID and display modal 
“Text: This app is using data webhooks. It will only be visible to Reapit Customers on AWS

**Logic**
Should check if webshooks are active =’true’
Does not include application.install and or application.uninstall. 
Customer ID is not ‘SBOX’. 

**Example:** 
  {
      "id": "08d94d02-57f6-46c7-891f-7a83ac068813",
      "created": "2021-07-22T11:17:51",
      "modified": "2021-08-11T10:23:48",
      "applicationId": "XXXX",
      "url": "https://URL.com",
      "description": "",
      "topicIds": [
        "appointments.created"
      ],
      "customerIds": [
        "KEA",
        "DAH"
      ],
      "active": true,
      "ignoreEtagOnlyChanges": true

- If no webhooks are available or only application.install and or application.uninstall or all inactive: 
Text: “This app is not using data webhooks and is visible to ALL users”

HollyJoyPhillips:
@willmcvay Thanks for the listed and private options, works really well. 

I've tested the webhooks check button but it doesn't seem to be working as expected. 

See 'Test Holly July' app has a data webhook (Applicant Modified) but on the Admin portal it's still showing for 'all customers' where is should only be for AWS customers: 

App ID 2fa8316b-cf70-44c9-a42d-104e2821cb73

![image](https://user-images.githubusercontent.com/52408607/130199822-85a431e4-b499-438e-9fbb-d632f81f263b.png)

![image](https://user-images.githubusercontent.com/52408607/130199881-f4734a6c-17d8-413b-a7aa-b870acb09a23.png)

Issue #4522, opened 2021-07-06
Add 'Notification Email' field on developer organisation 


HollyJoyPhillips:
**Summary**: When a developer organisation is setup an email address is provided. Any notifications (customer installs etc) are then sent to that email address by default. However, it has been requested that an alternative email address should receive those notifications instead. 

**Location:** https://developers.reapit.cloud/settings/organisation

**Task**: 
- Add an additional field on the organisation page 'Notification email' 
- New field 'notificationsEmail' available in the /developers endpoint 
- Add email validation (correct format etc)
- Field is not required 
- Helper text "TBC'

**Relates to:** https://github.com/reapit/foundations/issues/4100

HollyJoyPhillips:
@willmcvay Thanks for adding the field, just a couple of points (I should have added it to the ticket, apologies): 

- Please can you swap the Company Registration number' with the new email field (so it's directly under the 'Office Email'). 
- Also, can you call the field 'Notification Email' and use the text as helper (sub) text please? 

![image](https://user-images.githubusercontent.com/52408607/130090696-2513dfb2-5ac3-43b2-8fa3-190525d9a63e.png)

PR #4826, opened 2021-08-20
feat: #4522 #4766 admin dev fixes

willmcvay:
# Pull request checklist

fixes: #4522 
fixes: #4766


joshbalfour (approved):

packages/admin-portal/src/components/pages/apps-management/check-aws-button.tsx
```diff
@@ -25,7 +25,9 @@ export const checkIsAws = (subscriptions: WebhookModel[]): boolean => {
   const filtered = subscriptions.filter((subscription) => {
     const { topicIds, customerIds, active } = subscription
     const hasDataTopics = Boolean(topicIds?.filter((topic: string) => !INSTALL_TOPICS.includes(topic)).length)
-    const hasCustomerSubs = Boolean(customerIds?.filter((id: string) => !FILTERED_CLIENT_IDS.includes(id)).length)
+    const hasCustomerSubs = Boolean(
+      !customerIds?.length || customerIds?.filter((id: string) => !FILTERED_CLIENT_IDS.includes(id)).length,
+    )
```
bashleigh, line 30: this is hurting my brain a bit. Probably me as it's currently mush from all those aws issues. But here you're checking if length exists? `length >= 1` as `0` would result in truthy. and >= 1 would be false? If you just want to check if customerIds is not an array `!Array.isArray(customerIds)` would make more sense.

The second part is filtering the array and making sure the ids exist in the const? and checking there is at least one that exists? If so there's a function called `some` which might make more sense https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some
Some will return true if there is at least one present (not sure if it breaks once found) `customerIds.some(id => FILTERED_CLIENT_IDS.includes(id))`

This is all providing I've read this correctly?

bashleigh (changes requested):
Just a suggestion on readability :) otherwise looks good!

packages/admin-portal/src/components/pages/apps-management/check-aws-button.tsx
bashleigh: update: I didn't read it correctly

willmcvay: It's a bit weird - basically if the array is empty, the webhooks apply to all customers. If it's not empty, I want to filter it to see if it has length when SBOX has been removed. Worth looking at the test cases as it's a bit more obvious what's going on...

bashleigh (approved):

Issue #4589, opened 2021-07-15
Websocket implementation


bashleigh:
Create authenticated websockets on AWS for pipeline and release updates. This will enable us to display realtime info and responses about a pipeline's execution and their task exec logs

PR #4847, opened 2021-08-24
feat: #4589 websockets with pusher

bashleigh:
# Pull request checklist

fixes: #4589


packages/deployment-service/src/functions/pipeline/version-deploy.ts
```diff
@@ -20,10 +20,26 @@ export const versionDeploy: SQSHandler = async (event: SQSEvent, context: Contex
       const deployTaskIndex = pipelineRunner.tasks?.findIndex((task) => task.functionName === 'DEPLOY')
 
       // TODO check status
-      if (deployTaskIndex === -1 || typeof deployTaskIndex === 'undefined') {
+      if (deployTaskIndex === -1 || typeof deployTaskIndex === 'undefined' || !pipelineRunner.tasks) {
```
willmcvay, line 23: Can you not do `if (!deployTaskIndex || !pipelineRunner.tasks) {` ?

willmcvay (approved):
Just a small thing, can tidy up later 👍🏻

packages/deployment-service/src/functions/pipeline/version-deploy.ts
bashleigh: type checking as deployTaskIndex is a return from findIndex resulting in half spaceship result and possibly 'undefined' types. This condition only accepts 0 and above and doesn't treat 0 as false boolean.

Issue #4773, opened 2021-08-13
apps can be ejected into react code


joshbalfour:
**Background context or User story:**
As a developer who wants to add some features that the app builder doesn't support, I want to be able to "eject" the app into a create-react-app-esque typescript(?) project so I can add my fancy stuff.

**Specification or Acceptance Criteria:**
- each page is exported as a separate file
- components are available in a components directory with the view of common components being available via npm
- the code should be reasonably human-readable

PR #4960, opened 2021-09-10
feat: #4773 eject to react code

joshbalfour:
# Pull request checklist

fixes: #4773


packages/app-builder-backend/src/eject/templates/app.ts
```diff
@@ -0,0 +1,49 @@
+import { js } from './js'
+import { lint } from './format'
+
+export const generateApp = () => {
```
willmcvay, line 4: Feels a bit fragile / hard to maintain. I think in the future, we should do our own CRA template similar to the foundations one that has all this stuff pre-built

willmcvay: Not for today though!

packages/app-builder-backend/src/eject/templates/packageJson.ts
```diff
@@ -0,0 +1,43 @@
...
+        graphql: '^15.5.1',
+        'react-router': '^5.2.0',
+        'react-router-dom': '^5.2.0',
+        typescript: '^4.4.2',
+
+        'omit-deep': '^0.3.0',
+        'query-string': '^7.0.1',
+        'clone-deep': '^4.0.1',
+
+        '@reapit/connect-session': '^3.0.0', // TODO: should update dynamically to latest version
```
willmcvay, line 22: Can you just use @latest ? we flag our NPM releases

willmcvay (approved):
There's a lot in there but looks good to me!

packages/app-builder-backend/src/eject/templates/packageJson.ts
joshbalfour: unfortunately when you do that it comes back with this :/ 
image

Issue #4856, opened 2021-08-26
Add properties as a listable object


joshbalfour:
**Background context or User story:**
As a user I should be able to list and search Properties as well as Books and Authors.

**Specification or Acceptance Criteria:**
- search properties
- get property

PR #4987, opened 2021-09-14
feat: #4856 property listing works

joshbalfour:
# Pull request checklist

fixes: #4856


packages/app-builder-backend/src/resolvers/property-resolver.ts
```diff
@@ -0,0 +1,87 @@
+import { gql } from 'apollo-server-core'
+import { Resolver, Query, Ctx } from 'type-graphql'
+import fetch from 'node-fetch'
+import { Property, PropertyFragment } from '../entities/property'
+import { PropertyModelPagedResult } from '../../../foundations-ts-definitions/types'
+import { Context } from '@/types'
+import { graphqlUri } from '../config.json'
+import { DocumentNode } from 'graphql'
+
+export const authors = [
```
willmcvay, line 10: Does this still need to be here now?

willmcvay (approved):
LGTM 👍🏻

Issue #4877, opened 2021-08-27
Add Keys API to GraphQL service


willmcvay:
**Background context or User story:**

We have a new Keys API on the platform - this needs to be added to the GraphQL schema, following the normal patterns when the API is finalised_

PR #5032, opened 2021-09-17
feat: #4877 graphql keys api

joshbalfour:
# Pull request checklist

fixes: #4877

todo:
* tests
* rate limit handling


packages/graphql-server/src/resolvers/configurations/configurations.d.ts
```diff
@@ -14,6 +14,7 @@ export type ConfigurationType =
   | '​​tenancyTypes'
   | '​​vendorTypes'
   | '​​worksOrderTypes'
+  | 'keyTypes'
```
willmcvay, line 17: So I think because keys is part of properties, it should be an extension of the properties API itself. The aim has always been to be a 1:1 map of the main API schema so feel this might be going off piste!

willmcvay (changes requested):
My bad, I should have briefed you properly on where this one should live. Code looks good, probably just needs moving about a bit - sorry

willmcvay (approved):
LGTM 👍🏻

Issue #4912, opened 2021-09-02
Add keys to the CRUDable object list


joshbalfour:
**Background context or User story:**

**Specification or Acceptance Criteria:**
- get keys
- add key
- "move" key

PR #5160, opened 2021-10-04
feat: 4912 crudable keys

joshbalfour:
# Pull request checklist

fixes: #4912


packages/app-builder-backend/src/resolvers/key-resolver.ts
```diff
@@ -0,0 +1,244 @@
...
+  async checkOutKey(
+    @Arg('propertyId') propertyId: string,
+    @Arg('keyId') keyId: string,
+    @Arg('movement', () => KeyMovementModelInput) movement: KeyMovementModelInput,
+    @Ctx() ctx: Context,
+  ): Promise<Key> {
+    const { accessToken, idToken } = ctx
+    if (!accessToken || !idToken) {
+      throw new Error('unauthorized')
+    }
```
bashleigh, line 208: could you make this snippet reusable in any way? Just thinking if we changed the auth method it would take a lot of regex to find all the locations

bashleigh: Might also be helpful to throw this as a http error that lambdas like? I'm not 100% on this but I know they do like an object with a status property `return { status: 401, message: "Unauthorised" 
 }`

packages/app-builder-backend/src/utils/graphqlFetch.ts
```diff
@@ -0,0 +1,41 @@
...
+    throw new Error(json.message)
+  }
+  if (json.data) {
+    return json.data[dataKey] as T
+  }
+  if (json.errors) {
+    throw new Error(json.errors[0].message)
+  }
+  throw new Error('Invalid response')
+}
```
bashleigh, line 41: EOF

packages/app-builder/src/components/hooks/objects/use-sub-objects.ts
```diff
@@ -0,0 +1,33 @@
...
+    const { args } = list
+    const isForThis = args.some((arg) => arg.name === `${lowercaseFirstLetter(typeName)}Id`)
+    const hasOtherRequiredArgs =
+      args.filter((arg) => !(arg.name === `${lowercaseFirstLetter(typeName)}Id` || context[arg.name])).length === 0
+
+    return isForThis && hasOtherRequiredArgs
+  })
+
+  return {
+    data: subobjects,
```
bashleigh, line 29:
```suggestion
  const subObjects = data.filter((type) => {
    const { list } = type
    if (!list) return false
    const { args } = list
    const isForThis = args.some((arg) => arg.name === `${lowercaseFirstLetter(typeName)}Id`)
    const hasOtherRequiredArgs =
      args.filter((arg) => !(arg.name === `${lowercaseFirstLetter(typeName)}Id` || context[arg.name])).length === 0

    return isForThis && hasOtherRequiredArgs
  })

  return {
    data: subObjects,
```

packages/app-builder/src/components/hooks/objects/use-sub-objects.ts
```diff
@@ -0,0 +1,33 @@
+import { useIntrospection } from '../use-introspection'
+import { usePageId } from '../use-page-id'
+
+export const lowercaseFirstLetter = (string: string) => string.charAt(0).toLowerCase() + string.slice(1)
+
+export const useSubObjects = (typeName?: string) => {
```
bashleigh, line 6: Add a return type here? I can see false, {data: T[], error, loading}? and true? so `SubObject | boolean`? Not entrely sure

packages/app-builder/src/components/hooks/use-introspection/query-generators.ts
```diff
@@ -13,37 +13,62 @@ import {
...
+  return {
+    query: gql`
+      ${listQuery}
+    `,
+    args,
+  }
 }
 
-export const getGetQuery = (queries: Array<QueryableField>, queryableObjectTypes: Array<IntrospectionObjectType>) => {
+export const getGetQuery = (
```
bashleigh, line 47: getGetQuery? :p

bashleigh (commented):
Only some small issues, the one I'm feeling anal about is the error throws for lambda returns (if they are lambda funcs)

packages/app-builder-backend/src/utils/graphqlFetch.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const getGqlString = (doc: DocumentNode) => doc.loc && doc.loc.source.body
+
+export const query = async <T>(
+  query: DocumentNode,
+  variables: Record<string, any>,
+  dataKey: string,
+  { idToken, accessToken }: AuthTokens,
+) => {
+  const res = await fetch(graphqlUri || 'https://graphql.dev.paas.reapit.cloud/graphql', {
```
willmcvay, line 18: Would be good if this came from the config.json

packages/app-builder/src/components/ui/user/ejectable/table.tsx
```diff
@@ -49,62 +71,122 @@ const DeleteButton = ({ disabled, typeName, id }: { disabled?: boolean; typeName
...
+            label: `View ${uppercaseSentence(object.name)}s`,
+          })),
+        ]
 
-        return {
-          cells: [
-            ...cells,
-            {
-              label: 'Edit',
+        cells = [
```
willmcvay, line 120: Just a general comment but the table logic is a bit hard to follow. Feels like it could get broken down a bit into some functions or similar. Just a thought though...

willmcvay (commented):
Just a couple of small comments from me

packages/app-builder-backend/src/resolvers/key-resolver.ts
joshbalfour: good call, done it the native way now with decorators 👍

packages/app-builder/src/components/hooks/use-introspection/query-generators.ts
joshbalfour: it be like that sometimes 😭

packages/app-builder-backend/src/utils/graphqlFetch.ts
joshbalfour: it does, i'm just providing a default for types sake

packages/app-builder/src/components/ui/user/ejectable/table.tsx
```diff
@@ -49,26 +71,91 @@ const DeleteButton = ({ disabled, typeName, id }: { disabled?: boolean; typeName
...
+    .map(([label, value]) => ({
+      label: uppercaseSentence(label),
+      value: (typeof value === 'object' ? undefined : value) as string,
+      children: typeof value === 'object' ? <ObjectTableCell obj={value} /> : undefined,
+      narrowTable: {
+        showLabel: true,
+      },
+    }))
+
+const getAdditionalCells = (
```
willmcvay, line 97: Yeah, much better, easier on the brain!

willmcvay (approved):
I'm happy if Ashleigh is :-)

bashleigh (approved):

PR #5344, opened 2021-11-02
fix: authentication animation

bashleigh:
# Pull request checklist

fixes: #


packages/utils-react/src/key-animation/__styles__/key-animation.ts
```diff
@@ -69,7 +72,7 @@ export const keyhole = css``
 export const keyholeMask = css`
   width: 2.5rem;
   height: 3.5rem;
-  background: var(--color-dark-blue);
+  background: #262f68;
```
bashleigh, line 75: some reason the var didn't work here making the mask invisible 🤷‍♀️  not sure why but this resolves the problem

joshbalfour (approved):

packages/utils-react/src/key-animation/__styles__/key-animation.ts
willmcvay: You got the variable wrong!
Screenshot 2021-11-02 at 12 40 41

willmcvay: Referring to your docs 😂

bashleigh: 🙄 typical 😂

bashleigh: Screenshot 2021-11-02 at 13 27 07
Still doesn't work!

PR #5421, opened 2021-11-08
feat: 5360 app builder apps are actual apps

joshbalfour:
# Pull request checklist

fixes: #

### DIFF
--- a/packages/app-builder-backend/src/platform/apps.ts
+++ b/packages/app-builder-backend/src/platform/apps.ts
@@ -1,6 +1,13 @@
 import fetch from 'node-fetch'
 import config from '../config.json'
-import { AppSummaryModelPagedResult } from '@reapit/foundations-ts-definitions'
+import {
+  AppSummaryModelPagedResult,
+  CreateAppModel,
+  ProblemDetails,
+  AppDetailModel,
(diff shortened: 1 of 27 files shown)


packages/app-builder-backend/src/platform/apps.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  return data.data
+}
+
+export const createMarketplaceAppRevision = async (
+  appId: string,
+  appRevision: CreateAppRevisionModel,
+  accessToken: string,
+) => {
+  const res = await fetch(`${platformApiUrl}/marketplace/apps/${appId}/revisions`, {
```
willmcvay, line 34: The utils-common package exports fetchers that handle all the fetch logic including returning the location headers etc. No point repeating etc https://github.com/reapit/foundations/blob/master/packages/utils-common/src/fetcher/index.ts

willmcvay (approved):
Happy to merge, will leave my comment up to you.

packages/app-builder-backend/src/platform/apps.ts
joshbalfour: cheers for the pointer, using fetcher now!

Issue #5556, opened 2021-11-23
The release dev workflow has a number of errors


willmcvay:
**Background context or User story:**

_See https://github.com/reapit/foundations/runs/4285631042?check_suite_focus=true errors in GQL, API key and Payments service release commands_

PR #5612, opened 2021-11-30
fix: #5556 various fixes

joshbalfour:
# Pull request checklist

fixes: #5556


packages/utils-node/package.json
```diff
@@ -24,6 +24,7 @@
     "commit": "concurrently \"yarn test --coverage --no-cache --silent --forceExit --detectOpenHandles --runInBand --watch=false && jest-coverage-badges --input src/tests/coverage/coverage-summary.json --output src/tests/badges\" \"yarn lint\""
   },
   "dependencies": {
+    "@homeservenow/serverless-aws-handler": "^0.2.4",
```
bashleigh, line 27: what do you need this for in here?

bashleigh (changes requested):
Looks good to me, just wondering do you need the serverless package on the node-utils?

packages/utils-node/package.json
joshbalfour: it's used here https://github.com/reapit/foundations/blob/master/packages/utils-node/src/authorisation/authorised.ts#L1 no idea why this was able to build before 😂

bashleigh (approved):

packages/api-key-service/src/functions/create-api-key.ts
```diff
@@ -21,7 +21,7 @@ export const createApiKey = httpHandler<ApiKeyDto, ApiKeyModel>({
       if (typeof customer === 'undefined' || !customer.developerId) {
         throw new Error('Unauthorised')
       }
-    } catch (e) {
+    } catch (e: any) {
```
willmcvay, line 24: Rather than `any` you could do 

`const error = e as Error` and then pass `error.message` to the exception handler? Maybe a bit nicer?

willmcvay (approved):
LGTM 👍🏻

PR #5678, opened 2021-12-10
fix: app builder backend deployment

joshbalfour:


packages/ts-bundler/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@reapit/ts-bundler",
+  "main": "src/ts-bundler.ts",
+  "private": true,
```
willmcvay, line 4: I think you might need to add the normal build, test etc commands but echoed out - not 100% but think some of the GH workflows might fail otherwise

joshbalfour: good catch!

willmcvay (approved):
One small thing, it might be alright - worth just checking the workflow file commads before merging. Nice one aside from that 👍🏻

PR #5928, opened 2022-01-20
feat: auto package bump release

bashleigh:
# Pull request checklist

fixes: #

replaces #5245


joshbalfour (approved):

packages/auto-bump/src/app.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  const workingPackage = JSON.parse(workingPackageRaw)
+  workingPackage.version = version
+
+  await fs.promises.writeFile(fileName, `${JSON.stringify(workingPackage, null, 2)}\n`)
+}
+
+export default (app) => {
+  app.onAny(async (event) => {
```
bashleigh, line 21: I don't like this bit, couldn't get `release.published` to work which was the event that should've been triggered (and was when debugging) just wouldn't call the function given to it 🤷‍♀️ so settled for onAny as the action handles the event for us

Issue #4013, opened 2021-04-21
Provide Typescript definitions for Webhook payload


shaneosullivan:
**Is your feature request related to a problem? Please describe.**
When consuming webhooks in Node, it is currently required to use the "ping" feature and record the resulting payload as examples to build off of. 

**Describe the solution you'd like**
It would be better to have well defined Typescript definitions for each of the payload types, so we do not have to write them ourselves.

**Describe alternatives you've considered**
I've looked at each of the example webhook payloads from the "ping" tool and written my own Typescript types. These may or may not be accurate, as they are each based off a single example payload, which as far as I know could have multiple different shapes.  However there is no way to tell currently as the docs are not comprehensive, so I'll find out when a partner causes a bug live in my production.

**Additional context**
None

willmcvay:
Hi @shaneosullivan see: https://foundations-documentation.reapit.cloud/api/webhooks#example-payload our webhooks post the same object interface as per the existing definitions i.e. for the contacts.modified webhook, the old and new keys will be a ContactModel, in the ts-definitions package. This would apply to all entities i.e. property.created would be PropertyModel and so on.

For the webhook payload itself and for the couple of webhooks not part of the platform eg application.install, happy to add this to the backlog and we'll look at providing definitions in the future. For now, hopefully you should be able to construct your own interface from the docs.

shaneosullivan:
Thanks - I'm using this for now:

```

export type ReapitTopic =
  | "contacts.modified"
  | "contacts.created"
  | "contacts.modified"
  | "landlords.created"
  | "landlords.modified"
  | "propertyimages.created"
  | "propertyimages.modified"
  | "properties.created"
  | "companies.created"
  | "companies.modified"
  | "application.install"
  | "application.uninstall";

export interface ReapitPayload {
  eventId: string;
  entityId: string;
  customerId: string;
  eventTime: string; // ISO8601
  topicId: ReapitTopic;
}

// This is not included in Reapit's Typescript definitions for some reason.
interface ReapitApplicationModel {
  customerId: string;
  customerName: string;
  applicationId: string;
  applicationName: string;
  customerAddress: AddressModel;
  customerEmail: string;
  organisationId: string;
}

export interface ReapitCompanyCreatedPayload extends ReapitPayload {
  new?: CompanyModel;
}

export interface ReapitCompanyModifiedPayload
  extends ReapitCompanyCreatedPayload {
  old?: CompanyModel;
}

[...]

PR #5942, opened 2022-01-21
feat: #4013 webhook types

bashleigh:
# Pull request checklist

fixes: #4013 

## Usage

```ts

const handleEvent = (event: ReapitWebhookPayloadType) => {
  switch (event.topicId) {
    case reapitWebhookTopicEnum.APPLICATION_INSTALL:
      // event.new is AppDetailsModel
      // event.diff is Partial<AppDetailsModel>
      myParticularEvent(event)
      break;
    default:
      //event.new is any
  }
}

const myParticularEvent = (event: ReapitWebhookApplicationInstallEvent) => {
  // event.new is AppDetailsModel
}

```


packages/foundations-ts-definitions/types/webhook-schema.ts
```diff
@@ -0,0 +1,306 @@
+import {
```
willmcvay, line 1: So you will need to move this to it's own directory like the traffic-schema. Reason being the types dir gets rm -rf ed by the definitions cron job so only the auto generated types can live in the types directory

packages/foundations-ts-definitions/types/index.ts
```diff
@@ -7,3 +7,4 @@ export * from './marketplace-schema'
 export * from './payments-schema'
 // @ts-ignore
 export * from './platform-schema'
+export * from './webhook-schema'
```
willmcvay, line 10: See below, you need to remove this line and add the webhook schema here https://github.com/reapit/foundations/blob/master/packages/foundations-ts-definitions/scripts/create-index-file.js#L19

willmcvay (changes requested):
Sorry it's a bit janky but you just need to tweak a couple of bits, can you also add that usage thing you added to the PR to the README.md

Other than that LGTM.

packages/foundations-ts-definitions/types/webhook-schema.ts
bashleigh: thought this was the case. Forgot to check!

willmcvay (approved):

Issue #5885, opened 2022-01-19
There should be a "CRA" for Node apps


willmcvay:
**Background context or User story:**

_There should be a quickstart scaffolder for simple Node services within the mono-repo_

**Specification or Acceptance Criteria:**

- See #3523 we need a similar scaffolder so we can get started with Node packages quickly and with our standard config
- May use Yeoman but open to other CLI tools, see #3523 
- Should have CDK, ESLint, Jest and other tooling like config manager out the box
- Should deploy to Lambda and have a local dev server
- Open to choice of framework (express, nest, fastify, ANOther), depending on group agreement
- Does not need to work as a standalone NPM package _yet_ however, the ultimate goal is to be able to offer node apps to deploy to Troy so give some consideration as to how this might work.

PR #5979, opened 2022-01-26
feat: #5885 cna

bashleigh:
# Pull request checklist

fixes: #5885


packages/create-node-app-generator/app/templates/express/_config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "port": 3000
```
willmcvay, line 2: Wouldn't the base config need connectSecret and client id? We should be instantiating auth somewhere right? https://foundations-documentation.reapit.cloud/app-development/connect-session#node-usage

bashleigh: Will we always need this by default you think? Could use api-key

Issue #6114, opened 2022-02-09
The useReapitUpdate hook displays a success message when a request fails


willmcvay:
**Background context or User story:**

_When using useReapitUpdate, if a success message is specified and an update throws a 400, the message is shown and the hook returns a success update_

**Specification or Acceptance Criteria:**

- Should show the error message and fail gracefully

PR #6137, opened 2022-02-11
fix: #6114 handle reapit errors better

willmcvay:
# Pull request checklist

fixes: #6114

### DIFF
--- a/packages/utils-react/src/use-reapit-data/use-reapit-update.ts
+++ b/packages/utils-react/src/use-reapit-data/use-reapit-update.ts
@@ -1,7 +1,8 @@
+import { logger } from '@reapit/utils-react'
 import { useState, useEffect } from 'react'
 import { useAsyncState } from '../use-async-state'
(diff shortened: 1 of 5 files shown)


packages/utils-react/src/use-reapit-data/use-reapit-update.ts
```diff
@@ -95,6 +96,11 @@ export const send =
       let location
       let fetchResponse
 
+      if (!response.ok) {
+        const errorRes = await response.json()
+        throw new Error(handleReapitError(errorRes ?? {}))
```
bashleigh, line 101: setError?

bashleigh (approved):

packages/utils-react/src/use-reapit-data/use-reapit-update.ts
willmcvay: It does it in the catch block :-)

Issue #5915, opened 2022-01-20
Table Data UI should support the removal of non-required fields


willmcvay:
**Background context or User story:**

_When rendering a shared entity in a table, I should be able to customise the fields that are viewed_

**Specification or Acceptance Criteria:**

- If a field is not required in a shared entity, I should be able to remove it from the table list view, as I can do with form fields

PR #6185, opened 2022-02-17
feat: 5915 customisable table columns

joshbalfour:
# Pull request checklist

fixes: #5915


packages/app-builder/src/components/ui/user/form.tsx
```diff
@@ -20,6 +20,12 @@ const defaultProps = {
   destination: '/',
 }
 
+const strToCamel = (str: string) =>
```
willmcvay, line 23: Worth breaking into a util? You use in a couple of places...

PR #6269, opened 2022-03-02
feat: cdk watch

joshbalfour:
Workflow to use this should be:

1. Change the name of your stack to be `${stackName}-${developersName}` so `app-builder-backend` becomes `app-builder-backend-josh` for example
2. Run `yarn release:watch` in your package directory
3. Make changes, view logs from lambda executions in your terminal
4. When done, run `yarn release:destroy` to destroy your temporary stack
5. Revert the stack name change from 1.

If we make a lot of use of this these can probably be automated at some point


willmcvay (approved):
Can you add the usage instructions to the bundler README please? Other than that LGTM 👍🏻

joshbalfour:
@bashleigh heads up: I changed the commands in deployment-service and moved the `yarn build` and `yarn bundle` steps from `yarn release` so that they could be executed inside the stack https://github.com/reapit/foundations/pull/6269/files#diff-fa992979bd6c806d4cf0b3405ca34472f0d35db9f7ca362d6746838496e796b7

Issue #6457, opened 2022-03-23
The pipelines page should have it's own secondary nav item


willmcvay:
**Background context or User story:**

_Currently the pipelines page is a tab on the edit app form. It should be it's own item in the `/apps/:id` page_

**Specification or Acceptance Criteria:**

- Should be feature flagged via a whitelist of developer ids held in the config.json, similar to how we handle live chat

PR #6488, opened 2022-03-25
feat: #6457 pipeline navigation refactor and updatable

bashleigh:
# Pull request checklist

fixes: #6457


joshbalfour (approved):
LGTM 👍

packages/developer-portal/src/types/global.d.ts
```diff
@@ -17,6 +17,7 @@ export type Config = {
   swaggerUri: string
   PUSHER_KEY: string
   DEPLOYMENT_SERVICE_HOST: string
+  pipelineWhitelist?: string[]
```
willmcvay, line 20: Should be required field, you need to add it to the /core/index.tsx too where the config gets instantiated

willmcvay (changes requested):
Just the config thing if you don't mind :-)

packages/developer-portal/src/types/global.d.ts
bashleigh: oh ok, and force a default value of [] 👍🏼

willmcvay: Yep, nice one

willmcvay (approved):

PR #6528, opened 2022-03-31
feat: iaas menu item location

bashleigh:
# Pull request checklist

fixes: #6530


packages/developer-portal/src/components/pages/iaas/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+} from '@reapit/elements'
+import { PipelineModelInterface } from '@reapit/foundations-ts-definitions'
+import { ApiNames, GetActionNames, getActions } from '@reapit/utils-common'
+import { useReapitGet } from '@reapit/utils-react'
+import React, { useEffect, useState } from 'react'
+import { useHistory } from 'react-router'
+import { navigate } from '../../../utils/navigation'
+import { cardCursor } from '../apps/list/__styles__'
+
+const PipelineCard = ({ pipeline }: { pipeline: PipelineModelInterface }) => {
```
willmcvay, line 26: Can you break into a separate component, add a props type and tests please

packages/developer-portal/src/components/pages/iaas/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  meta: {
+    currentPage: number
+    itemCount: number
+    itemsPerPage: number
+    totalItems: number
+    totalPages: number
+  }
+}
+
+export const IaaS = () => {
```
willmcvay, line 74: Can you add the FC type please

packages/developer-portal/src/components/pages/iaas/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    action: getActions(window.reapit.config.appEnv)[GetActionNames.paginatePipeline],
+    headers: {
+      Authorization: connectSession?.idToken as string,
+    },
+    fetchWhenTrue: [connectSession?.idToken],
+    queryParams: {
+      page,
+    },
+  })
+  useEffect(() => setPagination(pipelines || undefined), [pipelines])
```
willmcvay, line 90: Seems a bit odd to set undefined if pipelines are falsy. Why not make the setPagination type to be `useState<Pagination<PipelineModelInterface> | null>` then you can set pagination with pipelines without the `||`

packages/developer-portal/src/components/pages/iaas/index.tsx
```diff
@@ -0,0 +1,125 @@
+import Routes from '@/constants/routes'
```
willmcvay, line 1: File needs tests

packages/developer-portal/src/utils/pipeline-helpers.ts
```diff
@@ -0,0 +1,42 @@
+import { Intent } from '@reapit/elements'
```
willmcvay, line 1: File needs tests please

packages/elements/src/components/card/card-components.tsx
```diff
@@ -60,10 +60,10 @@ export interface CardContextMenuProps extends HTMLAttributes<HTMLDivElement> {
 export interface CardProps extends HTMLAttributes<HTMLDivElement> {
   hasMainCard?: boolean // Should we render an Image card as the main card?
   // Supplied text options for the various fields
-  mainCardHeading?: string
+  mainCardHeading?: string | ReactElement
```
willmcvay, line 63: `string | ReactElement` better expressed as `ReactNode` as that covers anything that can be rendered eg numbers etc

packages/developer-portal/src/components/pages/iaas/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    queryParams: {
+      page,
+    },
+  })
+  useEffect(() => setPagination(pipelines || undefined), [pipelines])
+  useEffect(() => refreshPagination(), [page])
+
+  return (
+    <>
+      <Title>IaaS</Title>
```
willmcvay, line 95: All pages have a `SecondaryNavContainer` on the LHS of the page for controls and so on. Can you add in? See here for the markup https://github.com/reapit/foundations/blob/master/packages/developer-portal/src/components/pages/elements/index.tsx#L19

willmcvay: FYI you might need to add the IAAS route to this to make it render nicely https://github.com/reapit/foundations/blob/master/packages/developer-portal/src/constants/pages.ts#L3

Will get removed when the dev portal is fully refactored

willmcvay (changes requested):

packages/elements/src/components/card/card-components.tsx
bashleigh: that's the one

willmcvay (approved):
LGTM 👍🏻

PR #6608, opened 2022-04-12
[WIP] refactor: deployment service nestjs

bashleigh:
# Pull request checklist

fixes: #


joshbalfour (dismissed):

packages/deployment-service/.env.dist
```diff
@@ -0,0 +1,4 @@
+MYSQL_ROOT_PASSWORD=root
```
willmcvay, line 1: Erm, I think this should be gitignored. And probably best to change the PW

packages/deployment-service/src/default-header-interceptor.ts
```diff
@@ -0,0 +1,14 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common'
+import { Observable } from 'rxjs'
+import { Response } from 'express'
+
+@Injectable()
+export class DefaultHeaderInterceptor implements NestInterceptor {
+  intercept(context: ExecutionContext, next: CallHandler): Observable<any> {
+    const response = context.switchToHttp().getResponse<Response>()
+
+    response.setHeader('Access-Control-Allow-Origin', '*')
```
willmcvay, line 10: Think you can just use the cors package to fix this in Sonarcloud

willmcvay (changes requested):
I've had a good look but TBH I'm not familiar enough with Nest to offer any meaningful feedback on the code, suggest maybe asking @joshbalfour to check it out as well before merging. Please can we fix the Sonarcloud issues and that DB creds leak too

packages/deployment-service/.env.dist
bashleigh: This is fine, it's a dist so the idea is to copy this from `.env.dist` to `.env` and use `.env` for dev

packages/deployment-service/src/default-header-interceptor.ts
bashleigh: I'd rather do it like this than to add another package

willmcvay (approved):

Issue #6767, opened 2022-05-05
The developer portal should support configuring environment variables


willmcvay:
**Background context or User story:**

_See #6300 there should be a new tab in the pipelines page that allows users to save and update environment variables for their app. Triggering a build should retrieve the variables and include them in the bundled app. To be compliant with the CRA template, they should be written and read from the `/src/reapit-config.json` file_

PR #6820, opened 2022-05-09
feat: #6767 parameter store for codebuild

bashleigh:
# Pull request checklist

fixes: #6767


packages/developer-portal/src/components/apps/pipeline/pipeline-environment.tsx
```diff
@@ -0,0 +1,188 @@
...
+                      Update
+                    </Button>
+                  </TableCell>
+                </TableRow>
+              ))}
+            </>
+          )}
+        </Table>
+      </div>
+      <Modal isOpen={isInserting} onModalClose={() => setIsInserting(false)} className={cx(elMb6, elMt6)}>
```
willmcvay, line 151: So the pattern for forms with tables is to put the form in the slide down component. Forms in Modals, generally Ana regards as a pattern to avoid if poss. If you use the Table React shorthand, it make this process much easier.

willmcvay (changes requested):
Looks good, can you just move the Modal markup for adding / removing variables into the table slidedown please

willmcvay (approved):

PR #7020, opened 2022-06-07
fix: pusher developer id of pipeline

bashleigh:
# Pull request checklist

fixes: #


packages/developer-portal/src/components/apps/pipeline/pipeline-deployments-table.tsx
```diff
@@ -75,11 +76,11 @@ export const handleNewRunner =
     }
   }
 
-export const PipelineDeploymentTable: FC = () => {
+export const PipelineDeploymentTable: FC<{ channel?: Channel & PresenceChannel }> = ({ channel }) => {
```
willmcvay, line 79: Can you define this as an interface so you don't need to use any as the channel type elsewhere?

willmcvay (approved):

Issue #6775, opened 2022-05-05
App builder should have a configurable nav component


willmcvay:
**Background context or User story:**

_Navigation is missing from the app-builder component lib leaving the user lost when creating an app. We should add the standard Elements nav as a component and allow users to configure the nav items from the config panel based on their app routes_

PR #7121, opened 2022-06-21
feat: app builder designed nav

joshbalfour:
# Pull request checklist

fixes: #6775


willmcvay (approved):
Question; I can't see a NavStateProvider or MediaStateProvider - usually you will get an error if these aren't supplied eg https://github.com/reapit/foundations/blob/master/packages/developer-portal/src/core/app.tsx#L24 Did you omit or is there a workaround?

joshbalfour:
@willmcvay they were already there from the template (i never removed them previously) https://github.com/reapit/foundations/blob/master/packages/app-builder/src/core/app.tsx#L16-L22

Issue #7223, opened 2022-07-06
AppMarket installed apps page


willmcvay:
**Background context or User story:**

_As per the Figma designs, rebuild the installed apps page_

PR #7264, opened 2022-07-12
feat: #7223 adds installed apps page

willmcvay:
# Pull request checklist

fixes: #7223

### DIFF
--- a/packages/foundations-ts-definitions/scripts/create-index-file.js
+++ b/packages/foundations-ts-definitions/scripts/create-index-file.js
@@ -17,6 +17,8 @@ module.exports = () => {
         '// @ts-ignore\n' +
         'export * from "../api-key-schema"' +
         '// @ts-ignore\n' +
+        'export * from "../marketplace-cms"' +
+        '// @ts-ignore\n' +
         'export * from "../webhook-schema"' +
         '// @ts-ignore\n' +
         files
--- a/packages/foundations-ts-definitions/types/index.ts
+++ b/packages/foundations-ts-definitions/types/index.ts
@@ -3,6 +3,7 @@ export * from '../traffic-schema/marketplace-traffic-event-schema' // @ts-ignore
 export * from '../deployment-schema' // @ts-ignore
 export * from '../api-key-schema' // @ts-ignore
 export * from '../webhook-schema' // @ts-ignore
+export * from '../marketplace-cms' // @ts-ignore
 export * from './marketplace-schema'
 // @ts-ignore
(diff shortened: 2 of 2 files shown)


packages/app-market/src/components/apps-installed/apps-installed-suggested.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return 3
+}
+
+export const handleSortConfigs = (appsBrowseConfigState: AppsBrowseConfigCollection | null) => () => {
+  const featuredHeroApps =
+    appsBrowseConfigState?.data.filter((config) => config.configType === 'featuredHeroApps') ?? []
+  const heroApps = appsBrowseConfigState?.data.filter((config) => config.configType === 'heroApps') ?? []
+
+  const featuredHeroAppDetails = featuredHeroApps
+    .map(({ content, filters }) => {
```
bashleigh, line 31: you can use reduce to minimize null returns 

```ts
featured = featuredHeroApps.reduce((array, { content, filters }) => {
  if (content?. imageUrl && Array.isArray(filters.id) && filters.id.length >= 1) {
     array.push({ imageUrl: content.imageUrl, id: filters.id[0], })
  }
}, [])
```

Reasons: in your method, there's 3 loops, the initial map, the flat and the filter func.

packages/app-market/src/components/apps-installed/apps-installed-suggested.tsx
```diff
@@ -0,0 +1,79 @@
...
+          imageUrl: content.imageUrl,
+          id: filters.id[0],
+        }
+      }
+    })
+    .flat()
+    .filter(isTruthy)
+
+  const heroAppDetails = heroApps
+    .map(({ content, filters }) => {
```
bashleigh, line 43: same again, please see above

bashleigh (approved):
Approved but with a suggestion on minimizing loops

joshbalfour (approved):

packages/app-market/src/components/apps-installed/apps-installed-suggested.tsx
willmcvay: I tried that, firstly there are only 2 items in the array plus it's memoised so not expensive.

Also, none of the above typechecks, to actually make it work you have to type cast all over the place, set unhelpful defaults and things that will make the result less reliable.

In the end I chose reducing the risk of runtime errors over a small performance gain given we are only iterating over two items.

Issue #7635, opened 2022-09-02
Initial high level refactor of Admin Portal


willmcvay:
**Background context or User story:**

_As part of the Admin Portal refactor, we can do some high level refactoring while we await the full specification_

**Specification or Acceptance Criteria:**

- [x] Replace menu with Elements 3 version
- [x] Replace Error boundary with new Sentry one (see App Market) 
- [x] Replace Login with Elements 3 version (see App Market)
- [x] Update Private route wrapper to new container structure for Elements 3
- [ ] Create a HOC to wrap components to be hidden when user is on the admin whitelist (readonly)
- [ ] Remove any obvious references to elements legacy in the core directory and pages we know will be kept i.e. Apps / Approvals / Developers / Usage / Billing
- [ ] Where possible replace any redux fetching code with Reapit Hooks and delete the old redux references on the above 5 pages to be retained

PR #7668, opened 2022-09-09
feat: #7635 initial admin refactor

willmcvay:
# Pull request checklist

fixes: #7635


packages/admin-portal/src/utils/navigation.ts
```diff
@@ -0,0 +1,19 @@
...
+import { MouseEvent } from 'react'
+
+export enum ExternalPages {
+  platformAPIDocs = 'https://foundations-documentation.reapit.cloud/api/api-documentation',
+}
+
+export const openNewPage = (uri: ExternalPages | string) => (event?: MouseEvent) => {
+  event?.preventDefault()
+  event?.stopPropagation()
+  window.open(uri, '_blank')
```
joshbalfour, line 11: if this is for third party websites we could add no opener so they don't have access to our page, and no referer so they don't know the user was previously on our page
```suggestion
  window.open(uri, '_blank', 'noopener,noreferrer')
```

joshbalfour (approved):
LGTM 👍  just one tiny suggestion i'm not 100% sure on

packages/admin-portal/src/utils/navigation.ts
willmcvay: Updated, cool 👍🏻

Issue #7750, opened 2022-09-21
Toggle option- Admin Portal


HollyJoyPhillips:
**Summary:** New feature for toggling in the Admin Portal dev-management as discussed. 

Should apply 0.00 
Speak to HP for more information if required. 

**Endpoint**: PUT /Developers/{ID}

Issue #7722, opened 2022-09-16
Admin approval diff should include all new added features


willmcvay:
**Background context or User story:**

_There are a number of missing fields in the admin portal revisions diff. Review this diffing tool and ensure all fields are available in the diff_

**Specification or Acceptance Criteria:**

- All fields changed should appear in the approval diff

Issue #6161, opened 2022-02-14
There should be an infra section in the Admin portal


willmcvay:
**Background context or User story:**

_I should be able to perform IAAS management as an admin in the portal_

**Specification or Acceptance Criteria:**

- Add new IAAS top level menu item
- Should render a searchable table by developer
- Should show basic details and have a teardown function as a CTA

PR #7759, opened 2022-09-22
feat: #6161 adds iaas admin

willmcvay:
# Pull request checklist

fixes: #6161 
fixes: #7722 
fixes: #7750


packages/admin-portal/src/components/iaas/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+              >
+                Delete
+              </Button>
+            </ButtonGroup>
+          </Modal>
+          {pipelines && (
+            <Pagination
+              currentPage={pipelines.meta.currentPage ?? 1}
+              numberPages={pipelines.meta.totalPages ?? 1}
+              callback={setPage}
```
bashleigh, line 189: think you need a useEffect on setPage to reload data?

packages/admin-portal/src/components/iaas/utils.ts
```diff
@@ -0,0 +1,57 @@
+import { Intent } from '@reapit/elements'
+import { PipelineRunnerType } from '@reapit/foundations-ts-definitions'
+
+export const buildStatusToIntent = (status: string): Intent => {
```
bashleigh, line 4: can we SRP these please? Looks like they've been duplicated? Best to keep them in one place so changes will be reflected everywhere

bashleigh (changes requested):
Looks good! Just 2 suggestions, one I think can be solved later when in use but the other one might cause a little headache for one of us later on. Just want to keep consistency between applications otherwise we'll endup with conflicts between the two if changes were to be made.

packages/admin-portal/src/components/iaas/index.tsx
willmcvay: No, it works correctly

packages/admin-portal/src/components/iaas/utils.ts
willmcvay: Yep, agreed, was just being lazy

bashleigh (approved):

PR #7760, opened 2022-09-22
fix: type error in CRA template

byt3sage:
…| { [K in keyof P]?: string | undefined; }'.'

# Pull request checklist

When generating a new app using the CRA template, you get an error when running `npm run start`
image

This PR attempts to resolve this (although it might be incorrect)

fixes: `Type '{} | P' does not satisfy the constraint '{} | { [K in keyof P]?: string | undefined; }'`. which is thrown from line 14 of a generated application (`const props: RouteComponentProps<P | {}> = {`)

I'm not a TS developer normally so this fix may be pointless or potentially highly incorrect.

### DIFF
--- a/packages/cra-template-foundations/template/src/core/__mocks__/mock-router.tsx
+++ b/packages/cra-template-foundations/template/src/core/__mocks__/mock-router.tsx
@@ -2,7 +2,13 @@ import { RouteComponentProps } from 'react-router'
 import { UnregisterCallback, Href } from 'history'
 
 // This is to mock out the dependencies for react router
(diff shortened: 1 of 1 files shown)


willmcvay (approved):
@jaetoole Thanks for this, we made exactly the same change in our code eg https://github.com/reapit/foundations/blob/master/packages/app-market/src/core/__mocks__/mock-router.tsx 

Don't worry about the workflows not passing, it's just because you are not a member of the repo. Will merge as safe, thanks for the contribution.

Issue #7883, opened 2022-10-10
Pipelines broken in both dev and prod


willmcvay:
**Background context or User story:**

_When visiting an app listing in the dev portal in both dev and prod I am seeing two separate errors._

**Specification or Acceptance Criteria:**

- In dev, I am seeing a 500 - bad gateway
- In prod, I am seeing a 403

PR #7904, opened 2022-10-12
fix: Pusher/auth call issues resolved

bashleigh:
# Pull request checklist

fixes: #7883

- Resolves content-type issue with pusher/auth call
- Removes dupication of pusher/auth call
- Fixes typeorm changes
- local cors issues
- Fixes creation issue where clientId is missing

### DIFF
--- a/packages/developer-portal/src/components/iaas/index.tsx
+++ b/packages/developer-portal/src/components/iaas/index.tsx
@@ -1,4 +1,6 @@
+import { COGNITO_HEADERS, URLS } from '@/constants/api'
 import { reapitConnectBrowserSession } from '@/core/connect-session'
+import { PusherProvider } from '@harelpls/use-pusher'
 import { useReapitConnect } from '@reapit/connect-session'
 import {
   SmallText,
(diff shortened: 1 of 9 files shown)


packages/developer-portal/src/components/iaas/index.tsx
```diff
@@ -74,47 +72,35 @@ export const IaaS: FC = () => {
...
-                      callback={(page) => {
-                        setPage(page)
-                      }}
-                    />
-                  )}
-                </div>
-              </>
-            )}
-          </PusherProvider>
+          <Title>Pipelines</Title>
```
willmcvay, line 75: The table row component has a useEvent listener. If we remove this provider it will stop working, or even maybe blow up right?

bashleigh: yea just noticed and rebased

willmcvay: I can't see how you resolved this, looks the same in the diff

bashleigh: my bad, I thougt there was a global provider for pusher

bashleigh: ma bad, I thought there was a global wrapper but it only applies to apps

willmcvay (approved):

PR #8004, opened 2022-10-27
feat: gql v2 improvements

bashleigh:
# Pull request checklist

fixes: #


packages/developer-portal/package.json
```diff
@@ -73,6 +75,7 @@
     "serverless-s3-deploy": "^0.10.1",
     "serverless-s3-remover": "^0.6.0",
     "serverless-single-page-app-plugin": "^1.0.4",
+    "supertest-graphql": "^1.1.4",
```
willmcvay, line 78: Shouldn't this be in the GQL package not Dev Portal?

packages/graphql-server-v2/src/server.ts
```diff
@@ -1,50 +1,61 @@
...
+      )
 
-    return result.data
-  } catch (e: any) {
-    console.error(e)
-    return {
-      error: e.message,
+      return result.data
+    } catch (e: any) {
+      console.error(e)
```
willmcvay, line 40: Can we add a CloudWatch logger in here?

packages/graphql-server-v2/src/server.ts
```diff
@@ -1,50 +1,61 @@
...
-  const swaggerResponse = await fetch(`${process.env.PLATFORM_API_BASE_URL}/docs/swagger/agencyCloud_swagger.json`)
-
-  const swagger = await swaggerResponse.json()
-  console.log('Starting GQL server')
 
+export const bootstrap = async (axiosInstance: AxiosInstance): Promise<Express> => {
   const schema = await createSchema({
-    swaggerSchema: swagger as any,
-    callBackend: handlePlatformCall,
+    swaggerSchema: `${config.PLATFORM_API_BASE_URL}/docs/swagger/agencyCloud_swagger.json`,
```
willmcvay, line 49: Is the plan still to generate this on a cron job rather than generating the schema on the fly each time?

packages/developer-portal/package.json
bashleigh: yes it should

packages/graphql-server-v2/src/server.ts
bashleigh: am not sure it's possible in the way that we planned, I was going to attempt a different method today. Currently, using the package I can route all resolvers to a singular function whereas generating the schema I'd have to bypass this package in the codebase and attempt to setup apollo etc

willmcvay (approved):
LGTM 👍🏻

packages/developer-portal/src/components/graphql/index.tsx
```diff
@@ -1,64 +1,13 @@
...
-  useEffect(handleLoadGQLPlayground(setLoaded), [])
-
-  useEffect(handleInitGQLPlayground(loaded, graphQlRef, connectSession), [graphQlRef, connectSession, loaded])
-
   return (
     <>
       <Title>GraphQL</Title>
-      {!loaded && <Loader />}
-      <div className={graphQLWrapper} ref={graphQlRef} />
+      {window.reapit.config.GRAPHQL_EXPLORER ? <GraphQLV2 /> : <GraphQLV1 />}
```
willmcvay, line 10: Why not just use the appEnv? As in `window.reapit.appEnv === 'production' ? "V1": "V2"`

packages/developer-portal/src/types/global.d.ts
```diff
@@ -18,6 +18,7 @@ export type Config = {
   DEPLOYMENT_SERVICE_HOST: string
   pipelineWhitelist: string[]
   swaggerWhitelist: string[]
+  GRAPHQL_EXPLORER: boolean
```
willmcvay, line 21: As above, then we don't need to worry about updating all the configs

packages/graphql-server-v2/src/server.ts
willmcvay: Is this still relevant?

packages/developer-portal/src/components/graphql/index.tsx
```diff
@@ -1,64 +1,13 @@
...
-  useEffect(handleLoadGQLPlayground(setLoaded), [])
-
-  useEffect(handleInitGQLPlayground(loaded, graphQlRef, connectSession), [graphQlRef, connectSession, loaded])
-
   return (
     <>
       <Title>GraphQL</Title>
-      {!loaded && <Loader />}
-      <div className={graphQLWrapper} ref={graphQlRef} />
+      {window.reapit.config.appEnv === 'development' ? <GraphQLV2 /> : <GraphQLV1 />}
```
willmcvay, line 10: `window.reapit.config.appEnv !== 'production'` or you will get v1 locally

Issue #8236, opened 2022-12-08
Spike: Update useReapitData hooks for React 19


willmcvay:
**Background context or User story:**

_The current way we use the Reapit Data hooks is going to cause some unwanted double rendering in React 19. Can see this if we switch on StrictMode in 18. We need to look at refactoring the update and get hooks, possibly with React Query or similar to avoid needless re-renders and simplify the code_

**Specification or Acceptance Criteria:**

- Clone reapit data hooks into own package
- Fix re-rendering issue by adding strict mode to an app and refactoring hooks
- If possible, maintain a similar, if not the same API to avoid refactoring pain
- Do a small experiment with a page in one prod app replacing the old hook with a new one so we can test in the wild

PR #8276, opened 2022-12-19
feat: #8236 reapit data refactor

willmcvay:
# Pull request checklist

fixes: #8236


joshbalfour (approved):
LGTM 👍

packages/use-reapit-data/package.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "dependencies": {
+    "@reapit/connect-session": "workspace:packages/connect-session",
+    "@reapit/foundations-ts-definitions": "workspace:packages/foundations-ts-definitions",
+    "@reapit/utils-common": "workspace:packages/utils-common",
+    "@tanstack/react-query": "^4.20.4",
+    "@tanstack/react-query-devtools": "^4.20.4",
+    "axios": "^1.2.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
```
bashleigh, line 33: probably be better in dev deps? Or at least just `react-dom`. Our apps will likely not require react-dom for this package

packages/use-reapit-data/src/use-reapit-get.ts
```diff
@@ -0,0 +1,74 @@
...
+  reapitConnectBrowserSession,
+  action,
+  queryParams,
+  uriParams,
+  headers,
+  fetchWhenTrue,
+  onSuccess,
+  onError,
+}: ReapitGetParams): ReapitGetState<DataType> => {
+  const { connectSession } = useReapitConnect(reapitConnectBrowserSession)
```
bashleigh, line 39: does having this call here cause duplicated token creation requests?

packages/use-reapit-data/src/use-reapit-get.ts
```diff
@@ -0,0 +1,74 @@
...
+      if (!onError) {
+        errorSnack(errorString)
+        logger(new Error(errorString), connectSession)
+      }
+    },
+    enabled: fetchWhenTrue?.length ? Boolean(fetchWhenTrue?.filter(Boolean).length) : true,
+  })
+
+  const result = data ? data : null
+  const errorString = error?.message ? error.message : null
```
bashleigh, line 71: Why null instead of undefined? Thinking about type checking as `!null` is not the same as `!undefined`. `!null` will return true as `null` is Object whereas undefined is undefined type

packages/use-reapit-data/src/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+  }, '')
+
+// Same principle for this function but where your list is an array of strings or numbers
+export const stringListToBatchQuery = (list: (string | number)[], queryKey: string): string =>
+  list.reduce((query: string, nextItem: string | number, index: number) => {
+    if (!index) {
+      return String(nextItem)
+    }
+    return `${query}&${queryKey}=${nextItem}`
+  }, '')
```
bashleigh, line 98: is this an issue with how .NET handles array query params? Does `key[0]=&key[1]=` not work? I remember having this problem but not how I solved it

packages/use-reapit-data/src/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+    action: getActions(window.reapit.config.appEnv)[GetActionNames.getTrafficStats],
+    queryParams: {
+      appId: listToBatchQuery<AppSummaryModel>(apps, 'id', 'appId')
+    },
+    fetchWhenTrue: [dateFrom, dateTo],
+  })
+
+   listToBatchQuery<AppSummaryModel>(apps, 'id', 'appId') === MOCK_ID1&appId=MOCK_ID2&appId=MOCK_ID3
+ * 
+ */
```
bashleigh, line 80: Still need this block?

packages/use-reapit-data/src/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+    : defaultMessage
+    ? defaultMessage
+    : 'An unknown error has occurred, please refresh the page and try again.'
+  const fieldErrors = errors?.map(({ field, message }) => `"${field}: ${message}"`)
+  const fieldString = fieldErrors ? fieldErrors.join(', ') : ''
+
+  return `${messageString} ${fieldString}`
+}
+
+export const getUrl = (action: GetAction | UpdateAction, queryParams?: Object, uriParams?: Object) => (): string => {
```
bashleigh, line 49: I have an idea on how we can make our uriParams and body types changed depending on the type of request. Will need a refactor though so probably not best to do in this PR. We'll have to talk about it and make a ticket

bashleigh (changes requested):
Looks good! Just added some comments and questions on minor points

packages/use-reapit-data/src/utils.ts
willmcvay: Yeah, so I spoke to Ryan about this. We both thought it should work fine but he couldn't guarantee it. It was a just in case util, I copied across from React Utils rather than a new thing. It's used in a few places, so didn't want to bin it and I want to be able to remove the old hooks from the React Utils package ultimately.

Going to leave in for now but we may well deprecate later

packages/use-reapit-data/src/utils.ts
willmcvay: Sure, happy to discuss if you want to suggest something. Main thing was just to have no breaking changes initially...

packages/use-reapit-data/package.json
willmcvay: I moved just React to peer dependencies

bashleigh (approved):

PR #8307, opened 2023-01-10
refactor: mjml before release

bashleigh:
# Pull request checklist

fixes: #


packages/cognito-custom-mail-lambda/package.json
```diff
@@ -7,8 +7,9 @@
     "dist"
   ],
   "scripts": {
+    "generate-html": "yarn ts-node src/scripts/generate.ts",
```
willmcvay, line 10: Why not just put this in the build step then you don't need to commit the html and you know you always have the latest compiled templates before releasing?

packages/cognito-custom-mail-lambda/src/mailer/templates/confirm-registration.html
```diff
@@ -0,0 +1,368 @@
+<!doctype html>
```
willmcvay, line 1: See my point above, you can then remove the html files

willmcvay (changes requested):

packages/cognito-custom-mail-lambda/package.json
bashleigh: yea guess so, needed them for tests 😭

willmcvay (approved):
LGTM

Issue #8413, opened 2023-01-27
Bug: Button with chevronRight, when in loading state, chevron is removed


bashleigh:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behaviour eg:
Using a chevron on a button and setting loading={true}. The chevron is removed

**Expected behaviour**

**Screenshots**
If applicable, add screenshots to help explain your problem. Please ensure that screenshots do not include sensitive data such as personally identifiable information. If you do have private or sensitive information you need to share that will help with aiding resolution of this issue, please email sinfo@reapitfoundations.zendesk.com adding the issue number as the subject line. Please only use this mailbox to share information you cannot add to this issue.

<img width="525" alt="Screenshot 2023-01-27 at 16 07 58" src="https://user-images.githubusercontent.com/11574250/215133869-ea36124a-30f3-4057-bdc6-dbe4857f7c1e.png">

**Device (please complete the following information for web issues):**
 - Device: [e.g. iPhone6, PC, Mac]
 - Browser [e.g. Chrome, Safari]

**Additional context**

PR #8414, opened 2023-01-27
fix: loading with chevron

bashleigh:
# Pull request checklist

fixes: #8413


packages/elements/src/components/button/__styles__/index.ts
```diff
@@ -51,6 +51,33 @@ export const elButtonSize4 = css`
   }
 `
 
+export const ElButtonLoader = styled.div`
+  @keyframes spinAround {
```
joshbalfour, line 55: Does linaria have a [keyframes](https://styled-components.com/docs/basics#animations) equivalent?
Just thinking here if someone else defines spinAround it'll override this one and vice versa.
If it doesn't then maybe we should centralise our animation definition

joshbalfour (approved):
One small thing, otherwise LGTM :D

packages/elements/src/components/button/__styles__/index.ts
bashleigh: fair point, I shall have a look in the docs

PR #8460, opened 2023-02-06
feat: add disable property and styles to toggle radio component

bashleigh:
# Pull request checklist

fixes: #


packages/elements/src/components/toggle/toggle.tsx
```diff
@@ -79,11 +80,15 @@ export const ToggleRadio: ToggleRadioWrapped = forwardRef(
               name={name}
               value={value}
               type="radio"
+              disabled={disabled}
```
willmcvay, line 83: Small point but you don't need to pass disabled here as it's already being passed by the ...rest operator

willmcvay (approved):
I am fine with this if Ana is :-)

packages/elements/src/components/toggle/toggle.tsx
bashleigh: just went to remove and I'm using the disabled prop elsewhere so unable to default to ...rest unless I change `htmlFor={!disabled ? id : undefined}` to `rest.disabled`?

willmcvay: Ahh, cool

Issue #7654, opened 2022-09-07
Update Elements to latest design system


willmcvay:
**Background context or User story:**

_There have been some updates to the Design system. We should use this as an opportunity to ensure that Elements is still fully compliant with the design_

**Specification or Acceptance Criteria:**

- See separate word doc provided by Ana

PR #8461, opened 2023-02-06
feat: #7654 elements updates

willmcvay:
# Pull request checklist

fixes: #7654


joshbalfour (approved):
LGTM 👍

packages/developer-portal/src/components/analytics/costs/download-resources-table.tsx
```diff
@@ -80,7 +80,7 @@ export const DownloadResourcesTable: FC<DownloadResourcesTableProps> = ({ billin
             ],
             ctaContent: {
               onClick: handleDownloadCSV(billing, error),
-              icon: 'downloadSystem',
+              icon: 'downloadSolidSystem',
```
bashleigh, line 83: Feel like this is a bit OTT but seeing as this is an existing icon (not sure when added) but would this constitute a breaking change and require a major version change so not to conflict with those currently using this icon? Unless it's recently added then we might get away with making it a minor change release

packages/elements/src/components/card/__styles__/index.ts
```diff
@@ -71,9 +71,7 @@ export const ElCardSubHeading = styled.h6`
   }
 `
 
-export const elCardSubHeadingAdditionalExpanded = css`
-  margin-bottom: 1.25rem;
-`
+export const elCardSubHeadingAdditionalExpanded = css``
```
bashleigh, line 74: can be removed? Not an element so presuming it's an additional condition state?

bashleigh (commented):
Only one comment, just wanted to ask about the icon and our semver handling. Otherwise looks good.

packages/elements/src/components/card/__styles__/index.ts
willmcvay: It's left in to avoid breaking changes in case the class is being used. It now no longer does anything as the component is expanded by default

packages/developer-portal/src/components/analytics/costs/download-resources-table.tsx
willmcvay: It is an existing icon, yes

Issue #6286, opened 2022-03-03
`reapit config` becomes `reapit login`


joshbalfour:
**Background context or User story:**
Basically a "magic link" login flow for the CLI:
- `reapit login` opens webapp in browser (and displays url as a fallback):
-  webapp sends user through reapit-connect flow
- shows success/error page (this can be a small lambda)
- success page runs `mutation setToken($sessionId: ID!, $token: String!)`
- CLI has been polling API in background: `query getToken($sessionId: ID!) { token }`
- CLI shows success

PR #8519, opened 2023-02-14
feat: #6286 reapit connect login with cli

bashleigh:
# Pull request checklist

fixes: #6282


willmcvay (changes requested):
LGTM 👍🏻

packages/cli/src/services/auth/login.ts
```diff
@@ -0,0 +1,175 @@
...
+import open from 'open'
+import { writeFileSync, promises, existsSync } from 'fs'
+import 'isomorphic-fetch'
+import decode from 'jwt-decode'
+import ora from 'ora'
+
+export class LoginService {
+  private server
+  private readonly connectOAuthUrl: string = 'https://connect.dev.paas.reapit.cloud' // TODO Change dev to env
+  private readonly port: number = 8080
```
joshbalfour, line 20: this port might not be free, you can use something like https://www.npmjs.com/package/find-free-port to find a spare one

packages/cli/src/commands/auth/login.ts
```diff
@@ -0,0 +1,28 @@
...
+  description: 'test command for logging in via reapit connect',
+})
+export class LoginCommand extends AbstractCommand {
+  async run(
+    @Param({
+      name: 'clientId',
+      description: 'clientId of your app',
+      required: true,
+    })
+    clientId: string,
```
joshbalfour, line 16: Is the goal to log in to the CLI so that we can do away with API keys? if so maybe the CLI should have its own client ID rather than using theirs?

packages/cli/src/services/auth/login.ts
```diff
@@ -0,0 +1,175 @@
...
+  private readonly connectUrl: string = `${this.connectOAuthUrl}/login`
+  private readonly redirect = `http://localhost:${this.port}`
+  protected connectSession?: ReapitConnectSession
+
+  static storageLocation: string = `${homeDir()}/.reapit-connect.json`
+
+  private oauthUrl = (clientId: string, authCode: string) =>
+    `${this.connectOAuthUrl}/token?grant_type=authorization_code&client_id=${clientId}&code=${authCode}&redirect_uri=${this.redirect}`
```
joshbalfour, line 28: how do we make sure that the redirect uri is whitelisted in RC? should we have a range that it can pick from (whichever's free) and then whitelist those?

joshbalfour (changes requested):
Great PoC! I might have misunderstood the goal of this PR, but there are also a couple of other things.

willmcvay (approved):
Nice

joshbalfour (approved):

Issue #8641, opened 2023-03-02
Alter QR code in MFA app to include additional information on label


plittlewood-rpt:
**Background context**
The MFA management app allows users to setup MFA on their Reapit Connect accounts by presenting a QR code that the user can scan with an authenticator app of their choice. Part of the data in the QR code determines the label assigned to the item that gets added to the respective authenticator app. This label currently reads "Reapit Connect." This is fine where a user only has one Reapit Connect account, but becomes problematic when they have multiple accounts, as it's impossible to tell which second-factor code is for which account. We should enhance the label to include the email address for eliminate this problem

**Specification**
- Package: rc-mfa-config
- Alter QR code data to present label in the following format (this makes it consistent with Ory's out of box label, which will drop later in the year): `Reapit Connect:example@email.com` where example@email.com is the email address of the using configuring MFA

HollyJoyPhillips:
@plittlewood-rpt tested in dev and it's showing the email addresses correctly.

![image](https://user-images.githubusercontent.com/52408607/231385323-2b16a430-4d7d-48c6-9562-0519281de2d6.png)

PR #8796, opened 2023-03-24
feat: #8641 include email in qr label

plittlewood-rpt:
# Pull request checklist

Alters format of QR code label generated in RC MFA App. Slight amendment to ticket spec as the LastPass authenticator app doesn't like the use of the colon character. 

Display in Google Authenticator:
image

Display in LastPass Authenticator:
image

Display in Microsoft Authenticator:
image

I'm not sure why Microsoft Authenticator is failing to deal with the space in the issuer correctly, but given all these apps allow the account to be renamed anyway, I think this is OK.

@joshbalfour if this goes in we'll need to match up the format in Ory to this one

fixes: #8641


packages/rc-mfa-config/src/components/home/qr-code-verify.tsx
```diff
@@ -104,7 +107,7 @@ export const QrCodeVerify: FC<QrCodeVerifyProps> = ({ refreshAuthenticators, qrC
     },
   })
 
-  useEffect(handleDrawQrCode(modalIsOpen, canvasRef, qrCode), [canvasRef, qrCode, modalIsOpen])
+  useEffect(handleDrawQrCode(modalIsOpen, canvasRef, qrLabel, qrCode), [canvasRef, qrCode, modalIsOpen])
```
willmcvay, line 110: You need to add connectSession to the dependency array of this useEffect or it will not be called when the value of qrLabel changes. Passing qrLabel will cause it to be called each render because that variable is defined each render so you need to pass in the object it is derived from ie connectSession.

willmcvay (changes requested):

packages/rc-mfa-config/src/components/home/qr-code-verify.tsx
plittlewood-rpt: Had a read up, makes sense I think. Thanks

packages/rc-mfa-config/src/components/home/qr-code-verify.tsx
```diff
@@ -44,11 +44,13 @@ export const handleDrawQrCode =
   (
     modalIsOpen: boolean,
     canvasRef: MutableRefObject<HTMLCanvasElement | null>,
+    qrLabel: string,
     qrCode?: CreateAuthenticatorReturnType,
   ) =>
   () => {
+    console.log(qrLabel)
```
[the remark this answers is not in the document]
plittlewood-rpt: Good spot, thanks

Issue #8366, opened 2023-01-19
The Payments services should be included in billing statistics


willmcvay:
**Background context or User story:**

_Currently the core payments service and the new payments proxy do not figure in billing stats. As part of GTM, we will need to assess and bill customers for usage_

**Specification or Acceptance Criteria:**

- Payments and payments proxy APIs should be covered by billing stats

PR #8809, opened 2023-03-28
feat: #8366 adds payments billing model

willmcvay:
# Pull request checklist

fixes: #8366

### DIFF
--- a/packages/payments-service/src/core/billing-interceptor.ts
+++ b/packages/payments-service/src/core/billing-interceptor.ts
@@ -1,0 +1,47 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common'
+import { Observable } from 'rxjs'
+import { Request } from 'express'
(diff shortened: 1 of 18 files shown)


packages/payments-service/src/core/billing-interceptor.ts
```diff
@@ -0,0 +1,47 @@
...
+        EndpointMethod: method,
+        EndpointVersion: 'latest',
+        CustomerId: headers['reapit-customer'],
+        RequestId: null,
+        TimeStamp: new Date().toISOString(),
+      }),
+      QueueUrl: `https://sqs.eu-west-2.amazonaws.com/${config.AWS_ACCOUNT_ID}/Platform_Billing_HttpTrafficEvents`,
+    }
+
+    sqs.sendMessage(message, (err) => {
```
joshbalfour, line 39: sending messages to sqs is quick but there is a non-zero chance that the lambda could terminate before the message is sent

PR #8949, opened 2023-04-27
fix: remove bearer prefix for auth

bashleigh:
# Pull request checklist

fixes: #


packages/utils-nest/src/auth/token-provider.ts
```diff
@@ -17,7 +17,7 @@ export class TokenProvider implements AuthProviderInterface<any> {
   async resolve(request: Request): Promise<LoginIdentity> {
     try {
       const authorization = request.headers?.authorization as string
-      const claim = await connectSessionVerifyDecodeIdTokenWithPublicKeys(authorization)
+      const claim = await connectSessionVerifyDecodeIdTokenWithPublicKeys(authorization.replace('Bearer ', ''))
```
willmcvay, line 20: Huh? Why is Bearer being passed at all for an id token?

willmcvay (approved):

Issue #8947, opened 2023-04-27
Couldn't use the yarn 3 workspace in the IaaS


ss-dimasm:
**Describe the bug**
We're using yarn 3 workspace to develop and maintain apps and want to integrate the development release with the IaaS, but we got a bad luck when installing one of the package's dependencies in the IaaS with yarn package manager, the IaaS pipeline keep failing at "Install" task. It run smoothly in my local or GH action CI but not with the IaaS. Seems the yarn version that used in the IaaS is not switched to yarn 3 workspace it still using the classic one when installing the dependencies.

**To Reproduce**
Steps to reproduce the behaviour eg:
1. Have an app in the mono repo that managed by yarn 3 workspace
2. Integrate with the IaaS
3. Hit "Deploy" button
4. See the unexpected result

**Expected behaviour**
Because we're using yarn 3 workspace to maintain the packages' dependencies, so the IaaS should support that version instead using the classic version especially for installing and building tasks.

**Additional context**
- yarn package manager version that we use is 3.1.1
- error logs from one of the app when we're trying to deploy it with the IaaS
[96283816-acf4-47de-9f4d-f03e06b5b2bc.gz](https://github.com/reapit/foundations/files/11340266/96283816-acf4-47de-9f4d-f03e06b5b2bc.gz)

PR #8961, opened 2023-04-28
fix: #8947 yarn 3 workspaces

bashleigh:
# Pull request checklist

fixes: #8947


packages/deployment-service/src/codebuild/codebuild-executor-workflow.ts
```diff
@@ -125,6 +125,19 @@ export class CodebuildExecutorWorkflow extends AbstractWorkflow<{
   }) {
     const params = await this.parameterProvider.obtainParameters(pipeline.id as string)
 
+    const setupCommands = [
+      'n install 14',
```
willmcvay, line 129: Can we use Node 18 please? See my prev on tech debt and Node EOL

packages/deployment-service/src/codebuild/codebuild-executor-workflow.ts
```diff
@@ -138,12 +151,7 @@ export class CodebuildExecutorWorkflow extends AbstractWorkflow<{
               nodejs: 12,
```
willmcvay, line 151: Also Node 12 here!

bashleigh: done

willmcvay (approved):

Issue #7351, opened 2022-07-22
Migrate Node Apps to CDK


willmcvay:
**Background context or User story:**

Currently we have both CDK and Serverless deployments. This is messy, ideally we should use only one build script and Serverless is limited as a framework, hence CDK

**Specification or Acceptance Criteria:**

- In turn, set up each of the labeled node apps to have their own CDK stack, inheriting from the base
- Should use the yarn release command
- Should use GraphQL as a template
- When deployed to dev, switch over the DNS record to the new stack by pointing at the new Gateway instance
- Delete the old stack using serverless remove
- When dev confirmed as working, repeat the process for production releases
- Delete the old serverless.yml and serverless packages

PR #9055, opened 2023-05-15
refactor: #7351 security lambda cdk

bashleigh:
Refactor security header lambda cdk deployment method

fixes: #9065


packages/ts-scripts/src/cdk/components/site.ts
```diff
@@ -88,33 +90,31 @@ export const createSite = async (
...
+      edgeLambdas: edgeLambda && [
+        {
+          functionVersion: edgeLambda.currentVersion,
+          eventType: LambdaEdgeEventType.ORIGIN_RESPONSE,
         },
-        behaviors: [{ isDefaultBehavior: true }],
-      },
-    ],
+      ],
+    },
```
bashleigh, line 101: Not sure if that problem you had with S3 will be effected by this? @willmcvay

willmcvay: Erm, not sure

willmcvay (approved):

Issue #9081, opened 2023-05-17
Update node version used in github workflows


bashleigh:
Update the node version to v18 that is used in github workflow actions

PR #9086, opened 2023-05-18
chore: updated node version for github actions

bashleigh:
# Pull request checklist

fixes: #9081


joshbalfour (approved):

willmcvay (commented):
So I have a feeling some of the underlying actions use old versions of Node eg ssh-agent, get-tag, we probably ought to update these registry actions too as long as it's not painful :-)

Issue #9144, opened 2023-05-24
Fix deployment-service deploy method


bashleigh:
Deployment service has a bug after updating packages, the zipped build is now too large, use tsup and custom bundler to reduce the size of the deployable zipped files.

Work partially done on this branch
chore/update-deployment-service-build-and-bundle-method

There is currently a runtime issue where a symbol from dynamodb data mapper is not working, likely because of the compiling from api-key-verify service. Potential fix could be to make this an externally configured package with tsup

PR #9194, opened 2023-06-01
chore: update deployment service build and bundle method

bashleigh:
Refactor deployment service to use tsup to avoid large payload rejections from AWS bundle.zip. 

fixes: #9144


bashleigh:
it lives <3

joshbalfour (approved):

PR #9299, opened 2023-06-19
feat: rc-mfa-config anz deployment

joshbalfour:
Adds the ANZ r53 zone which DevOps can then delegate to to put it live.
Once they have we can uncomment the other lines in order to add it to the cloudfront dist


packages/rc-mfa-config/cdk-stack.ts
```diff
@@ -1,19 +1,60 @@
-import { createBaseStack, createSite, getAccountId } from '@reapit/ts-scripts/src/cdk'
+import { Stack, createBaseStack, createSite, getAccountId } from '@reapit/ts-scripts/src/cdk'
 import { join } from 'path'
+import {
+  aws_route53 as route53,
+  aws_route53_targets as targets,
+  // aws_cloudfront as cloudfront,
+  // aws_certificatemanager as acm,
```
bashleigh, line 7: still need these commented lines?

Issue #10243, opened 2023-11-28
Reapit ConnectEmail template updates


plittlewood-rpt:
**Background context**
Update admin-user-invite, forgot-password and confirm-registration email templates in the custom mailer (and equivalent templates in the internal user service) as per designs from ANZ

PR #10245, opened 2023-11-28
feat: 10243 Update templates with designs from ANZ

plittlewood-rpt:
# Pull request checklist

Updates template format for Reapit Connect emails. To be used in all regions once we get the nod to roll them out

fixes: #10243

Forgot password example:
image

Admin user invite example:
image

Confirm account example:
image


willmcvay (approved):
LGTM

PR #10263, opened 2023-11-29
chore: Tweak message so it reads better and looks nicer

plittlewood-rpt:
# Pull request checklist

- Removes dodgy space in href
- Rework deletion message so it reads better

No ticket, just noticed it whilst I was testing something


plittlewood-rpt:
@willmcvay call me pedantic but this reads better now!

willmcvay (approved):

Issue #10485, opened 2023-12-21
Include submitting member email with App submissions/revisions


AshDeeming:
Admin portal 

Additional column on App revision details to include the Dev Org Member email of who submitted the revision

**Specification**
- Package: developer-admin-portal
- Include `createdBy` from AppRevisionModel in the table displayed in the frontend

PR #10588, opened 2024-01-15
feat: #10485 - Include Created By on incoming app revision details

RWilcox-Reapit:
Update to display the user who created the revision on the approvals screen. 
(This has been done in the drop-down as it's not until that point that the API call to /revisions is called)


plittlewood-rpt:
I think you've got an outdated snapshot in here somewhere - code looks OK though to me

plittlewood-rpt:
Hmm just looking at this again. The ideal is to add it to the table - there's nothing to really 'diff' with this field. Would a better solution be to add createBy to the response model that comes back from the API so it can be displayed on the table?

Issue #10421, opened 2023-12-14
New filters - RC Service App


HollyJoyPhillips:
**Task**: 
- Add option to filter by neg ID (agencyCloudId)

PR #10688, opened 2024-02-07
feat: #10421 neg id filter

bashleigh:
# Pull request checklist

fixes: #10421


packages/rc-service/src/components/users/index.tsx
```diff
@@ -114,6 +115,15 @@ export const UsersPage: FC = () => {
               <Label htmlFor="myId">Select User Group</Label>
             </InputGroup>
           </InputWrap>
+          <InputWrap>
+            <InputGroup
+              {...register('agencyCloudId')}
+              type="search"
+              icon="search"
+              placeholder="Search by Negotiator"
```
willmcvay, line 123: You wouldn't be searching by negotiator if the field is agencyCloudId

willmcvay (changes requested):

Issue #10709, opened 2024-02-09
Provide CSV on RC Service App


HollyJoyPhillips:
Provide a ‘Download CSV’ option on the RC Service App for 'Users' only
**Should include:** 
Name
Email Address
Company
Org ID
Neg ID
Date Created
First Login Date 
User Roles 
- Download option should respect the filters applied

PR #10722, opened 2024-02-12
feat: added method to download rows as csv

bashleigh:
# Pull request checklist

fixes: #10709 

> Small issue with page size, cannot fetch more than 999 per page. Not sure if there's a query param to increase it or whether I need to paginate the request to get all users for the download.


packages/rc-service/src/components/users/index.tsx
```diff
@@ -67,6 +73,61 @@ export const UsersPage: FC = () => {
...
+          {
+            type: 'text/plain;charset=utf-8',
+          },
+        )
+        fileSaver.saveAs(file)
+      }
+    } catch (error) {
+      logger(error as Error)
+    }
+    setCsvLoading(false)
```
willmcvay, line 128: Use a finally for this?

packages/rc-service/src/components/users/index.tsx
```diff
@@ -185,6 +246,11 @@ export const UsersPage: FC = () => {
       ) : users?._embedded?.length ? (
         <>
           <BodyText>Total Users: {users?.totalCount}</BodyText>
+          <ButtonGroup>
+            <Button intent="primary" loading={csvLoading} disabled={csvLoading} onClick={() => downloadAsCSV()}>
```
willmcvay, line 250: You can just pass downloadAsCSV to the on click right?

Issue #10942, opened 2024-03-21
Video Imagery - New App 


HollyJoyPhillips:
**Issue**: Placeholder image/video link is not rendering correctly: 
**Location**: https://developers.reapit.cloud/apps/new

Examples: 
https://developers.reapit.cloud/apps/new

![image](https://github.com/reapit/foundations/assets/52408607/a3fc43e5-c11c-4c8c-85bf-6e0db759e626)
Website Feed:
![image](https://github.com/reapit/foundations/assets/52408607/1a56a5d5-5cab-48c9-87c5-3da205266e3c)
Server-side: 
![image](https://github.com/reapit/foundations/assets/52408607/c85ac7d1-8520-4705-8ef0-4156463135da)

Location: https://developers.reapit.cloud/desktop

![image](https://github.com/reapit/foundations/assets/52408607/5a5b1ecd-e33f-4da7-94a3-9a4df2e73de4)

plittlewood-rpt:
@HollyJoyPhillips  this still needs releasing just FYI

PR #10996, opened 2024-04-05
fix: resolves issue with small video thumbnails in helper content

plittlewood-rpt:
# Pull request checklist

Resolves issue with tiny video thumbnail by forcing component to fill the available space. Issue was due to the svg element having a height of 1em when rendered. I'm not entirely sure where this was coming from, but the PR resolves the issue as shown below. The PR also alters one of the vite server properties which I had to adjust to get the application to run locally on my windows machine (using Ubuntu WSL). Without knowing the ecosystem very well I don't know if there'll be any adverse effects to changing this (I suspect not, given developers on Macs use localhost anyway), but please consider this change when reviewing.

image

fixes: #10942

### DIFF
--- a/packages/ts-scripts/src/vite/base-vite.js
+++ b/packages/ts-scripts/src/vite/base-vite.js
@@ -92,7 +92,7 @@ module.exports = (config, appName) =>
       alias: [{ find: '@', replacement: `${path.resolve(process.cwd(), 'src')}/` }],
     },
     server: {
-      host: true,
+      host: 'localhost',
       port: 8080,
     },
   }))
--- a/packages/developer-portal/src/components/apps/new/__tests__/__snapshots__/helper-content.test.tsx.snap
+++ b/packages/developer-portal/src/components/apps/new/__tests__/__snapshots__/helper-content.test.tsx.snap
@@ -35,6 +35,7 @@ exports[`HelperContent should match a snapshot 1`] = `
                 Every developer will have different use cases for our Platform and it is unlikely you will need to use all of the APIs and tooling we provide. However, the starting point for all integrations is to create an ‘App’,
               </mock-styled.p>
               <svg
+                height="100%"
                 width="100%"
               />
             </div>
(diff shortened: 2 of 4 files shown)


RWilcox-Reapit (approved):

plittlewood-rpt:
@willmcvay @bashleigh would one of you mind just giving me the nod with this - I feel like this might actually be a problem somewhere in Elements. This 'fixes' the issue as far as the Dev Portal goes, but feel like there might be more to it

packages/ts-scripts/src/vite/base-vite.js
```diff
@@ -92,7 +92,7 @@ module.exports = (config, appName) =>
       alias: [{ find: '@', replacement: `${path.resolve(process.cwd(), 'src')}/` }],
     },
     server: {
-      host: true,
+      host: 'localhost',
```
bashleigh, line 95: Not sure about this change? Did you have an issue with the config?

bashleigh (commented):
Have added a comment for now

willmcvay:
> @willmcvay @bashleigh would one of you mind just giving me the nod with this - I feel like this might actually be a problem somewhere in Elements. This 'fixes' the issue as far as the Dev Portal goes, but feel like there might be more to it

@plittlewood-rpt this will fix it - it's a custom SVG only used on that page. FYI the SVGR processor adds 1em sizing to all icons so they can be re-sized using fontSize, although height and width is fine too. Looks like we missed applying a size to that particular SVG - your fix is fine.

plittlewood-rpt:
Thanks - yes I caught the 1em thing when debugging it. Will merge this then. Thanks for checking over it

packages/ts-scripts/src/vite/base-vite.js
plittlewood-rpt: I did indeed. I had to change this so that it would run in a windows WSL environment.

Issue #10935, opened 2024-03-21
Connect Session Server throws error


willmcvay:
**Background context or User story:**

_Not sure how many people use it but the Connect Session server function is broken. Instantiating the `ReapitConnectServerSession` class in a simple node script throws an error "ReferenceError: self is not defined:" - my guess is one of the methods is not bound to "this"._

**Specification or Acceptance Criteria:**

- Should correctly instantiate the server session class

**Replication Steps**
- Use the https://github.com/reapit/foundations-code-examples/tree/main/node-website-feed project
- Upgrade the reapit connect session package and other dependencies to see the error

bashleigh:
Would you know their typescript/javascript setup? Sounds like a esbuild/ecmascript problem?

willmcvay:
@bashleigh  https://github.com/reapit/foundations-code-examples/tree/main/node-website-feed

I found the bug trying to update the dependencies in this simple Node example - upgrade connect session and you should see

bashleigh:
Thanks Will. Looking at that repo I think the problem is definately related to the build of the packages. I'd recommend we split the server package out into a seperate package and or with it's own seperate build method so we can include node and other requirements without the frontend related requirements

PR #11037, opened 2024-04-17
refactor: #10935 seperate node scripts from fe

bashleigh:
# Pull request checklist

fixes: #10935


packages/connect-session-server/tsup.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from 'tsup';
+import fs from 'fs';
+
+const pkgJson = JSON.parse(fs.readFileSync('package.json', 'utf-8'))
+
+export default defineConfig({
+    entry: ['src/index.ts'],
+    target: 'node16',
```
willmcvay, line 8: Node 18 right? Think 16 is out of support

packages/connect-session/README.md
```diff
@@ -198,49 +198,6 @@ The below example shows how to embed on any static or dynamic page with a single
...
-
-app.listen('3000', () => {
-  console.log('App is listening on 3000')
-})
-```
-
-As per the browser usage, you will need to instantiate the class with your initializers, in this case connectClientId, connectOAuthUrl (in the same way as the browser module), but with the addition of the connectClientSecret you obtain from your app listing page.
-
-The module will fetch and refresh your session as the token expires, caching it locally to minimise calls to Reapit Connect token endpoint.
+This has moved to [@reapit/connect-session-server](https://www.npmjs.com/package/@reapit/connect-session-server)
```
willmcvay, line 203: Doesn't look like you removed the code from the browser package

willmcvay (approved):

PR #11403, opened 2024-08-21
feat: #236 authorizer auth0 deployment service

bashleigh:
Deployment service authorization method. 

fixes: #236


packages/deployment-service/cdk/lib/cdk-stack.ts
```diff
@@ -188,6 +189,28 @@ export const createStack = async () => {
...
+    'deployment-service-authorizer-lambda',
+    path.resolve(__dirname, '..', '..', 'dist', 'authorizer'),
+    'authorizer/index.handler',
+    {
+      ISSUERS: config.ISSUERS,
+      CLIENT_ID: config.COGNITO_CLIENT_ID,
+      CONNECT_USER_POOL: config.CONNECT_USER_POOL,
+    },
+    undefined,
+    undefined,
```
plittlewood-rpt, line 203: Your default memory allocation is 1024Mb which I'd say for node is way too high. I'd consider explicitly setting this here to something lower (can't remember which undefined param it is). You'll more than likely see the same performance with 512

packages/utils-authorizer/src/index.ts
```diff
@@ -0,0 +1,91 @@
+import { APIGatewayTokenAuthorizerEvent, APIGatewayAuthorizerResult } from 'aws-lambda'
+import * as jwt from 'jsonwebtoken'
+import { JwtRsaVerifier, CognitoJwtVerifier } from 'aws-jwt-verify'
+
+export const authorizerHandler =
+  (customChallenge?: (event: APIGatewayTokenAuthorizerEvent, decodedToken: jwt.Jwt) => Promise<void | Error>) =>
+  async (event: APIGatewayTokenAuthorizerEvent): Promise<APIGatewayAuthorizerResult> => {
```
plittlewood-rpt, line 7: Some comments in here would be useful just so anybody looking at it can understand why it's doing what it's doing

plittlewood-rpt (changes requested):

plittlewood-rpt (approved):

PR #11404, opened 2024-08-21
feat: #236 authorizer auth0 gql server

bashleigh:
# Pull request checklist

fixes: #11394


packages/graphql-server/cdk-stack.ts
```diff
@@ -38,9 +39,34 @@ const createStack = () => {
...
+    },
+    undefined,
+    undefined,
+    undefined,
+    aws_lambda.Runtime.NODEJS_18_X,
+  )
+
+  const authorizer = new aws_apigateway.RequestAuthorizer(stack, 'graphql-server-authorizer', {
+    handler: authorizerLambda,
+    identitySources: [aws_apigateway.IdentitySource.header('authorization')],
```
plittlewood-rpt, line 65: We had an issue when looking at the payments service with a possible case sensitivity issue here. The docs don't explicitly call this out for authorizers specifically, but it does mention they are processed in a case sensitive way. See [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html). You might therefore want to build something here to deal with `Authorization` vs `authorization`

plittlewood-rpt (approved):

PR #11422, opened 2024-08-28
chore: update packages

bashleigh:
# Pull request checklist

fixes: #


plittlewood-rpt (approved):
I feel like I'm blindly approving this but as long as we're following whatever process we used to follow when doing the monthly housekeeping, then that's fine

plittlewood-rpt (changes requested):
CI is failing actually. Can you check that out first please

PR #11443, opened 2024-09-17
refactor: remove mfa management with button to my-account

bashleigh:
# Pull request checklist

fixes: RPLT -416


packages/rc-mfa-config/src/components/home/index.tsx
```diff
@@ -98,10 +101,23 @@ export const HomePage: FC = () => {
     },
   })
 
+  useEffect(() => {
```
plittlewood-rpt, line 104: For what its worth, I spoke to Will about this this morning (on my dev portal PR) and he pointed me towards rendering <Loader /> if token was null over useEffect. I don't know why, just wanted to flag that in case there's any specific reason to do it one way over another

bashleigh: I don't get it? Surely the token will never be null? We'll redirect away to /login because home is within a private route

packages/rc-mfa-config/src/constants/urls.ts
```diff
@@ -0,0 +1,11 @@
+export const URLS = {
+  production: {
+    reapitConnectMyAccount: 'https://connect.reapit.cloud/my-account',
```
joshbalfour, line 3: could generate these based on the token issuer rather than use static values (thinking about anz etc. here)

plittlewood-rpt: @joshbalfour I've updated this this morning. Would you mind giving this a once over again so we can get it in? Cheers

PR #11446, opened 2024-09-20
feat: rplt-397 signpost users to my-account spa when signing in via Auth0

plittlewood-rpt:
# Pull request checklist
- Updates DeveloperPortal password page to no longer permit password updates directly from the page, instead signposting user to my-account page
- Users with a token issue by cognito will continue to see the existing fields

image


packages/developer-portal/src/constants/urls.ts
```diff
@@ -0,0 +1,11 @@
+export const URLS = {
+  production: {
+    reapitConnectMyAccount: 'https://connect.reapit.cloud/my-account',
```
joshbalfour, line 3: Should we generate this from the issuer instead?

joshbalfour (approved):

joshbalfour (approved):

plittlewood-rpt:
Snyk issues can be ignored, mock tokens being picked up

PR #11449, opened 2024-09-23
chore: remove foundations-ts-definitions

bashleigh:
# Pull request checklist

fixes: #


willmcvay (approved):
Have approved however;

- Adding the `Platform` and `Marketplace` namespaces is fine - not against this per se however, it does create a significant breaking change. We should ensure we bump the version to 2.x to reflect this.
- Additionally, can you update the docs to reflect the breaking change - a "migrating from v1 to v2 guide. It's pretty simple to document, just don't want people to be confused. See https://foundations-documentation.reapit.cloud/app-development/foundations-ts-defintions#usage 

Cheers

plittlewood-rpt:
Cheers Will. We've gone to v2 anyway because of this and Ashleigh's got a documentation update in draft so will make sure that's all covered

PR #11462, opened 2024-10-11
feat: rplt-495 add notice about refresh token rotation

plittlewood-rpt:
- Adds notice on appropriate steps in app registration wizard about authorisation code flow apps getting single use refresh tokens after november 2024
- Adds modal that prompts when users hit the new-app page. We'll want to revert this change at some point

RC_RotatingRefreshTokens


plittlewood-rpt:
@willmcvay  @HollyJoyPhillips I was in two minds about the modal here, because right now it'll display every time someone hits the new app page. I think given the frequency at which developers hit this page it doesn't really matter, and we can just loop back round and remove it after say a month? Can you let me know your thoughts?

joshbalfour (approved):

PR #11468, opened 2024-10-15
fix: gql bundle method correction

bashleigh:
# Pull request checklist

fixes: #


bashleigh:
Doesn't work. There are several problems such as 
- ts-bundler doesn't include node_modules for some reason
- ts-bundler expects there to be paths defined in tsconfig (redundent because of the foundations-ts-definitions removal) can be resolved by removing the exception condition but point above persists
- refactoring to use tsup (removing ts-bundler) results in issue with graphql version exception? Haven't been able to understand it or how to resolve it

PR #11507, opened 2025-01-22
fix: 739 RPLT custom resource s3 oac repair

bashleigh:
RPLT 739

Seperated custom resources to update buckets and distros. 

- Firstly, run the policy fix for buckets. Doing this seperately before the OAC being added to the distros will prevent outages
- Secondly, the OAC custom resource will add the OAC to each existing distro avoiding rate limit exceptions by retrying when the rate limit is hit. Also waiting between calls
- Thirdly, the S3 permissions custom resource can be deployed, resulting in the buckets no longer being public


packages/deployment-service/src/resolve-production-s3-bucket-permissions.ts
```diff
@@ -0,0 +1,132 @@
...
+  Action: string[]
+  Condition?: string[]
+  Resource: string
+  Principal: {
+    [s: string]: string
+  }
+}
+
+const resolveBucket = async (client: S3Client, bucketName: string) => {
+  await client.send(
```
plittlewood-rpt, line 25: Discussed with AS: There will be a very short period during which Cloudfront won't be able to access the bucket because of the policy being dropped, however the policy HAS to be dropped to be able to alter the ACL settings. Given Cloudfront should cache the bucket contents anyway this is an acceptable, but noted risk

packages/deployment-service/cdk/lib/resolve-production-OAC-custom-resource.ts
```diff
@@ -12,7 +12,7 @@ export class ResolveProductionOACCustomResource extends Construct {
         handler: 'dist/resolve-production-apply-OAC-to-all-distros.resolveProductionApplyOACToAllDistros',
         code: aws_lambda.Code.fromAsset('bundle/resolve-production-apply-OAC-to-all-distros.zip'),
         memorySize: 512,
-        timeout: Duration.minutes(5),
+        timeout: Duration.minutes(10),
```
plittlewood-rpt, line 15: Probably increase this to 15 mins just to be safe then we've got the maximum time allocated

plittlewood-rpt (changes requested):

packages/deployment-service/src/resolve-production-s3-bucket-permissions.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const statements: BucketPolicyStatement[] = [
+    {
+      Effect: 'Allow',
+      Action: ['s3:Get*', 's3:List*', 's3:Put*'],
+      Resource: `${bucketArn(bucketName)}/*`,
+      Principal: {
+        AWS: '*',
+      },
+      // Condition?? TODO for PaaS account
```
joshbalfour, line 59: were we wanting to do this now?

packages/deployment-service/src/resolve-production-apply-OAC-to-all-distros.ts
```diff
@@ -3,10 +3,100 @@ import {
...
+      new ListDistributionsCommand({
+        Marker: marker,
+        MaxItems: 5,
+      }),
+    )
+
+    if (!fetchDistros.DistributionList.Items) return
+
+    // Await each call, avoids making 10 async calls to AWS to avoid rate limit
+    fetchDistros.DistributionList?.Items.forEach(async (distro) => {
```
joshbalfour, line 90: if the comment above is what the goal is, then you probably want a `for of` loop here https://stackoverflow.com/a/37576787

packages/deployment-service/src/resolve-production-s3-bucket-policies.ts
```diff
@@ -21,7 +12,11 @@ type BucketPolicyStatement = {
 
 const bucketArn = (bucketName: string): string => `arn:aws:s3:::${bucketName}`
 
-const resolveBucketPolicies = (client: S3Client) => async (bucketInputs: string[]) => {
+export const resolveProductionS3BucketPolicies = async () => {
+  const client = new S3Client({})
+
+  const bucketInputs = process.env.BUCKETS ? process.env.BUCKETS?.split(',') : []
```
joshbalfour, line 18: should we error here? i'm guessing we always need these to be set

packages/deployment-service/cdk/lib/resolve-production-S3-bucket-permissions-custom-resource.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const resourceProvider = new custom_resources.Provider(scope, 'resolve-production-S3-permissions', {
+      onEventHandler: resolveProductionS3Lambda,
+      logRetention: aws_logs.RetentionDays.TWO_WEEKS,
+    })
+
+    new CustomResource(scope, 'resolve-production-s3-permissions', {
+      serviceToken: resourceProvider.serviceToken,
+      properties: {
+        fistonly: true,
```
joshbalfour, line 74: 👊 what's this for?

packages/deployment-service/src/resolve-production-s3-bucket-policies.ts
```diff
@@ -21,7 +12,11 @@ type BucketPolicyStatement = {
 
 const bucketArn = (bucketName: string): string => `arn:aws:s3:::${bucketName}`
 
-const resolveBucketPolicies = (client: S3Client) => async (bucketInputs: string[]) => {
+export const resolveProductionS3BucketPolicies = async () => {
+  const client = new S3Client({})
```
joshbalfour, line 16: docs make it look like you need to specify it here but guessing this uses the region that the lambda's running in?

joshbalfour (commented):
few things i'm not 100% clear on, think just the async await issue is the most important

packages/deployment-service/cdk/lib/resolve-production-S3-bucket-permissions-custom-resource.ts
bashleigh: haha once only custom resource trigger

packages/deployment-service/src/resolve-production-s3-bucket-permissions.ts
bashleigh: I decided to do these changes in another PR once this has all been sorted. Wasn't confident with the conditionals as there's no types

plittlewood-rpt (approved):

PR #11510, opened 2025-01-28
feat: added form wizard and refactored register page

bashleigh:
## ~~Form Wizard~~

~~Created a reusable form wizard component to split forms into several paged/stepped forms. This component allows for implementation as follows~~

## Step Wizard

Instead of the previous implmentation of having a reusable form wizard with configurable forms which had an issue with outside controls. I've reimplementated the feature into a generic step wizard which is able to have contextualised controls outside of the root component. This allows for the registration form to be able to navigate to a specific step with an error message to be displayed.

Below is an example of the navigation happening and the error message being displayed
Screenshot 2025-02-17 at 13 16 29


bashleigh:
> what's the behaviour if the backend returns an error? are we able to link it to the respective field? if so, what happens when you're on the last step, hit submit, and there's an error related to a field on the first step?

As explained this morning, each form is validated when "next" is pressed. Meaning you cannot proceed to the next form unless the current form has been successfully passed validation. As for server side errors, there isn't an ability to navigate to the specific error or steps the errors occurred. Potentially we could solve this but reckon there will need to be a discussion as to what the best method or apporach would be visually and what would be possible technically

packages/developer-portal/src/components/form-wizard/form-wizard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const [formValues, setFormValues] = useState<TFieldValues>({} as TFieldValues)
+
+  const previousStep = () => {
+    setCurrentStep(currentStep - 1)
+  }
+
+  const nextOrSubmit: SubmitHandler<TFieldValues> = (values) => {
+    if (isSubmitStep) {
+      onSubmit(formValues)
+      return Promise.resolve(formValues)
```
plittlewood-rpt, line 26: What happens if the backend does return an error at this point (I know we've tried to marry up the validation rules so that shouldn't happen, but need to ensure its handled properly)

plittlewood-rpt (approved):

plittlewood-rpt:
@HollyJoyPhillips  please can you scheduled a design meeting to go through this. This is all a bit backwards. I don't know if the meeting with Ashleigh demoing it on her machine is better. As soon as this is merged we're blocked from doing anything with the developer portal

HollyJoyPhillips:
> @HollyJoyPhillips please can you scheduled a design meeting to go through this. This is all a bit backwards. I don't know if the meeting with Ashleigh demoing it on her machine is better. As soon as this is merged we're blocked from doing anything with the developer portal

Just to confirm, AS and I went through this last week on her machine and agreed on next steps before releasing to dev. Once additional changes have been made (e.g. handling how we navigate errors) we will have a follow up.

PR #11540, opened 2025-03-28
feat: rplt 864 app signout urls

bashleigh:
# Pull request checklist

fixes: #


packages/deployment-service/src/marketplace/marketplace.provider.ts
```diff
@@ -0,0 +1,85 @@
...
+   * @param domain The domain to be added to the app for login redirect/signout
+   */
+  async updateAppUrls(appId: string, domain: string, developerId: string, name: string): Promise<void> {
+    const { access_token, token_type } = await this.authenticate()
+
+    await firstValueFrom(
+      this.httpService.post<Marketplace.AppRevisionModel>(
+        `${this.config.url}/apps/${appId}/revisions`,
+        {
+          redirectUris: [`https://${domain}/login`],
```
joshbalfour, line 71: does this overwrite any existing ones?

joshbalfour (approved):
assuming adding a custom domain to a pipeline won't remove the current redirect/signout urls from the app?

packages/deployment-service/src/marketplace/marketplace.provider.ts
bashleigh: I've not been able to confirm. A revision is being created and needs to be approved, the request is a success however I can't see the urls In the app Details section. I 'think' the revision require approval but not sure where I can find that.

PR #11544, opened 2025-04-07
refactor: split http api function and prevent isolate calls

bashleigh:
# Split http api endpoints and prevent calls to none api endpoints from api lambda

rplt 942

### DIFF
--- a/packages/deployment-service/src/httpApi.ts
+++ b/packages/deployment-service/src/httpApi.ts
@@ -9,6 +9,7 @@ import { Server } from 'http'
 import express, { Express } from 'express'
 import * as bodyParser from 'body-parser'
 import { CorsHeaderInterceptor } from '@reapit/utils-nest'
+import { APIGatewayProxyResult } from 'aws-lambda/trigger/api-gateway-proxy'
 const crypto = require('crypto')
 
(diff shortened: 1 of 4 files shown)


packages/deployment-service/src/httpApi.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Handler, Context, APIGatewayEvent } from 'aws-lambda'
+import { eventContext } from 'aws-serverless-express/middleware'
+import { Server } from 'http'
+import express, { Express } from 'express'
+import * as bodyParser from 'body-parser'
+import { CorsHeaderInterceptor } from '@reapit/utils-nest'
+import { APIGatewayProxyResult } from 'aws-lambda/trigger/api-gateway-proxy'
+const crypto = require('crypto')
+
+global.crypto = crypto
```
joshbalfour, line 15: needed?

joshbalfour (approved):

packages/deployment-service/src/httpApi.ts
bashleigh: sort of? Will need to refactor the auth method tbh

PR #11563, opened 2025-04-16
fix: revert elements to 4.3.2

bashleigh:
elements ^4.3.2 when built causing problems with the snack function which results in navigation not working correctly

# Pull request checklist

fixes: #


plittlewood-rpt (approved):
lgtm. @willmcvay worth you being aware of this one for next week.

PR #11569, opened 2025-05-09
feat: rplt 996 dns pr

bashleigh:
## Wizard
Screenshot 2025-04-30 at 11 18 17

Screenshot 2025-05-12 at 11 02 36

## PR creation on devops-infrastructure-core-cdk

https://github.com/reapit-global/uk-devops-infrastructure-core-cdk/pull/1451
Screenshot 2025-05-12 at 10 26 16


packages/deployment-service/src/dns/devops.pr.provider.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      mockPipeline as PipelineEntity,
+    )
+
+    expect(mockGithubPullRequestProvider.obtainFile).toHaveBeenCalled()
+    expect(mockGithubPullRequestProvider.createPullRequest).toHaveBeenCalledWith(
+      expect.objectContaining({
+        PR: expect.objectContaining({
+          title: expect.stringContaining('cname-name'),
+          body: expect.stringContaining(`user that triggered this action: ${mockPipeline.dnsTrigger}`),
+          // .stringContaining('cname-name')
```
plittlewood-rpt, line 65: Remove commented code if we don't need it

packages/developer-portal/src/components/apps/pipeline/dns/dns-settings-page.tsx
```diff
@@ -57,6 +29,13 @@ export const DnsSettingsPage: FC<{
             notify your relevent Dev Ops team about the records mentioned below.
           </BodyText>
         </InputWrapFull>
+        <InputWrapFull>
```
plittlewood-rpt, line 32: Need to rethink this sentence:

`   Your custom DNS setup was completed on IaaS. For your custom domain to be working, you will now you need to
            notify your relevent Dev Ops team about the records mentioned below.`

The first part suggests it's already been set up. Maybe something like

`Setup a custom domain name for your deployment from this page. To be able to complete the setup, the relevant DevOps team will need to accept the records below. The pipeline will create a PR automatically for their review`

packages/deployment-service/src/dns/devops.pr.provider.ts
```diff
@@ -0,0 +1,89 @@
...
+        hosted_zone_identifier: 'reapit.cloud', // has to be reapit.cloud
+        ttl: 300,
+        records: [cname.value],
+        type: cname.type,
+      })
+    })
+
+    const updatedFile = dump(yml)
+
+    await this.githubPrProvider.createPullRequest({
```
plittlewood-rpt, line 60: We probably need some idea of when this fails. Can we catch a failure and generate some kind of alert off the back of it? Looks like we might risk silent failures here

packages/deployment-service/src/dns/devops.pr.provider.test.ts
bashleigh: I wanted to do multiple string contains but wasn't able to. Types suggested I could chain the functions but this resulted in "cannot call X of undefined"

packages/deployment-service/src/dns/devops.pr.provider.ts
bashleigh: We could? Can you spec that out for me please.

plittlewood-rpt (approved):

PR #11595, opened 2025-07-04
feat: rplt 1063 status change

bashleigh:
# Pull request checklist

fixes: #rplt 1063

## User Status Modal
image

## Status History 
image

There are a few outstanding items:
- [x] The Update button should not be clickable unless a Reason and Notes are populated
- [x] When the Reason is NOT Suspicious Activity, the "Send notifications" checkbox is disabled. The value should be reset when adjusting the Reason so it can't be left in a ticked and disabled state
- [x] The Status History table needs some work. The modal height should be fixed and the table should be scrollable. Column widths need adjusting


plittlewood-rpt:
@HollyJoyPhillips  please can you provide feedback on the UI. The tweaks we discussed have now been made. THanks

plittlewood-rpt (approved):

PR #11596, opened 2025-07-07
chore: upgrade aws-cdk for mysql version bump

bashleigh:
# Pull request checklist

fixes: #rplt-1060


packages/ts-scripts/src/cdk/components/rds-database.ts
```diff
@@ -10,7 +10,7 @@ export const createDatabase = (
   bastion: boolean = false,
 ): rds.DatabaseCluster => {
   const db = new rds.DatabaseCluster(stack, name, {
-    engine: rds.DatabaseClusterEngine.auroraMysql({ version: rds.AuroraMysqlEngineVersion.VER_3_05_2 }),
+    engine: rds.DatabaseClusterEngine.auroraMysql({ version: rds.AuroraMysqlEngineVersion.VER_3_08_0 }),
```
plittlewood-rpt, line 13: I thought I'd replied to this apologies 

Couple of things. I'd suggest going to 3.08.2 which is the current default version (see https://docs.aws.amazon.com/AmazonRDS/latest/AuroraMySQLReleaseNotes/AuroraMySQL.Updates.30Updates.html)

Secondly, what's going to happen when this gets released? Presumably AWS will do an in place upgrade which will take the server offline whilst it happens. Is a blue/green deployment more appropriate?

PR #11611, opened 2025-08-21
feat: rplt-1153 replace hard wired terms with embedded pdf

plittlewood-rpt:
- Replaces the hard coded copy of the terms and conditions with a rendered PDF version. This makes it much easier to handle future changes to terms and conditions as it will just be a case of updating the PDF
- Updates the date that we check to determine if the modal should be presented for users which have accepted a previous version of the terms and conditions (the new one becomes active on 01/09/2025)

@willmcvay I'd appreciate if you could give this a once over. It works well and I've made sure the terms can't be accepted if the PDF doesn't load as it's better that we don't allow this to happen from a legal standpoint.

We are just waiting for the final copy of the updated terms after which there'll be a tweak to the file name but wanted some early feedback in case there's any objection to this approach.

Demo:
DevPortal_TandCPDF


rpt-uk-github:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/reapit-open-source/pr-checks/722ae8a3-d68b-42f9-bea5-5e27a59772ee?refs=6e401760eee69026d58a0be1f6f48bfb6aaf2b5a&source=prChecksComment)

willmcvay (approved):
Please check that PDF just gets served from the assets dir by CloudFront and isn't bundled. Assuming so, LGTM

